PR #6, opened 2016-05-05
Make sure build commands work out of the box

ldionne:
Cloning via the git protocol doesn't work out of the box (I get `Permission denied (publickey).`), while using https always works. Also, I put all the commands in a single Markdown code block, which
is prettier.


HDembinski:
makes sense, thanks!

PR #17, opened 2016-06-18
Cleanup & Serialization

klemens-morgenstern:
Hi Hans,

I added comments with copyright and moved the serialization to a custom header and added a compile test for the latter (b2 only). I also removed all unneeded macros (was quite unreadable with the locally declared ones).

It now has a new header (boost/histogram/serialization) with the serialization defs.
Also boost/histogram.hpp was added for convience, i..e it include histogram.hpp and serialization.hpp.

Lastly, I changed fill_c and wfill_c to fill(boost::iterator_range). Passing to iterators would've been an invalid ambiguity (when passing int), so now it takes a range. Should be fine, since it's already a boost library.

LG,

Klemens

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -1,0 +1,24 @@
+# (C) Copyright 2016 Klemens D. Morgenstern
+# Distributed under the Boost Software License, Version 1.0. (See accompanying 
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt.)
+
+# See http://www.boost.org/libs/iostreams for documentation.
+
+project /boost/histogram : source-location ../src ;
+
+# For each library with either link to existing binary, or build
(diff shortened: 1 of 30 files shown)


klemens-morgenstern:
Ok, I also changed the size check in histogram to std::range_error. It is not a critical error, hence it should be able to handle it; the assert would cause the program to crash immediatly.

klemens-morgenstern:
Ouh, some doc changes moved in here accidentily - I'll remove them if you want to pull. I'll open another pull request for doc stuff that later.

klemens-morgenstern:
Alright, tests pass as in your master branch - the fifth python test errors for some reason unknown to me...

HDembinski:
we have to work a bit on the details (see email discussion), but i am going to include this to go forward. thanks :) !

PR #18, opened 2016-06-18
Doc

klemens-morgenstern:
Switched to quickbook: http://klemens-morgenstern.github.io/histogram/

Don't know if you'll use that. This already includes the changes in the code found in #17.


include/boost/histogram/detail/nstore.hpp
```diff
@@ -64,13 +69,13 @@ class nstore {
...
+    size_(o.size_),
+    depth_(o.depth_),
+    buffer_(o.buffer_)
   {
-    std::swap(size_, o.size_);
-    std::swap(depth_, o.depth_);
-    std::swap(buffer_, o.buffer_);
+	o.depth_ = d1;
+	o.size_ = 0;
+	o.buffer_ = static_cast<void*>(0);
```
klemens-morgenstern: Seems to me like it's done ;). It's actually a very minor thing (since the compiler would optimize that anyway), but you avoid a copy in std::swap.

HDembinski: You are right, my bad for not reading carefully! :)


On 6/20/16 2:18 PM, Klemens Morgenstern wrote:
>
> In include/boost/histogram/detail/nstore.hpp 
> <https://github.com/HDembinski/histogram/pull/18#discussion_r67739348>:
>
> >    {
> > -    std::swap(size_, o.size_);
> > -    std::swap(depth_, o.depth_);
> > -    std::swap(buffer_, o.buffer_);
> > +	o.depth_ = d1;
> > +	o.size_ = 0;
> > +	o.buffer_ = static_cast<void*>(0);
>
> Seems to me like it's done ;). It's actually a very minor thing (since 
> the compiler would optimize that anyway), but you avoid a copy in 
> std::swap.
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub 
> <https://github.com/HDembinski/histogram/pull/18/files/b798f18d105070863a9ccc1d0837288f53d992e9#r67739348>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/ACgnouA6CGkuZx7_4Rd5ZviPjkCnrxb9ks5qNtlagaJpZM4I5But>.
>

-- 
Hans Dembinski
https://github.com/HDembinski

Proposed for Boost: histogram
https://github.com/HDembinski/histogram

Automatic one-dimensional unfolding of data distributions
https://github.com/HDembinski/aru-unfolding

PR #20, opened 2016-06-20
turned histogram python into proper pyd

klemens-morgenstern:
Damit da auch eine .pyd rauspurzelt und keine lib.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -9,16 +9,16 @@ project /boost/histogram : source-location ../src ;
 # For each library with either link to existing binary, or build
 # a library from th
 
-import python ;
-
 lib boost_histogram : axis.cpp basic_histogram.cpp histogram.cpp nstore.cpp zero_suppression.cpp ;
 
-lib boost_pyhistogram : 
+import python ;
+
+python-extension histogram : 
                         python/axis.cpp 
                         python/basic_histogram.cpp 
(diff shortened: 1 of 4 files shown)


klemens-morgenstern:
Alright, I added the Iterator-version with disable if, and thus removed boost::iterator_range. Should be a seperate PR, but I think that is a small change enough to be here.

klemens-morgenstern:
The enable_if header moved between boost 1.55 and 1.61, that's why it won't work. I don't wanna fiddle around with your travis.yml, so maybe you could modify that to get a newer version (1.56 at the least) and then it will work. If you do so, I'll pull this into my branch and then we'll see if it works.

HDembinski:
travis is a bit behind with boost, unfortunately. they provide a way to pull in a precompiled boost1.55 package, but nothing more recent than that. We can decide to drop using the boost package, and compile boost ourselves, then there is no problem in using the latest version. Using the package is convenient because it cuts down the compilation time, but it not necessary.

klemens-morgenstern:
I know, funny enough that this is one version too early. The official boost library (and boost.process) pull the complete boost from github, so you know that the whole things works on the develop branch with all the other libraries before a release. But that causes horrible compile times.

Now, if I'd be able to get the python tests to run, the boost.build would automatically only compile the libraries you actually need, not all of it.

PR #40, opened 2017-12-12
make fails to find sync_code.py with just "../doc".

Axel-Naumann:
Fixes
```
python: can't open file '../doc/sync_code.py': [Errno 2] No such file or directory
```
when running `make` after `cmake -DBOOST_ROOT=/Users/axel/build/boost/inst/ ../src/build/`


HDembinski:
Vielen Dank fuer den Fix!

PR #81, opened 2018-09-15
Remove build Serialization step as unnecessary

mloskot:
Serialization will be built automatically as part of test which requires the library as a dependency.

Add b2 command to build and run the speed tests.

### DIFF
--- a/doc/build.qbk
+++ b/doc/build.qbk
@@ -25,7 +25,6 @@ If you want to build this library as part of the Boost distribution, do this.
     git clone https://github.com/HDembinski/histogram.git
     mv histogram $BOOST_ROOT/libs
     cd $BOOST_ROOT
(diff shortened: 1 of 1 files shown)


HDembinski:
The speed tests are not intended to be part of the CI tests, but I think it is cool if they can be build with b2.
However, you cannot run the speed tests without external libraries, therefore we should not list them in build.qbk or add a note that the speed tests require external libraries.

mloskot:
I've removed the command from the docs. Sorry, I've only run the standalone `speed_cpp.cpp` test, missed the other require the third-parties.

HDembinski:
Ok, merging! We could still just build the speed test.

mloskot:
Good idea, done in #83

PR #85, opened 2018-09-17
Move CMakeLists.txt to top-level directory of the library

mloskot:
Add missing copyright header to CMakeLists.txt.
Rename build directory to [tools](https://www.boost.org/development/requirements.html#Organization).
Update Travis CI and AppVeyor scripts.
Ignore variants of build directory.
Ignore CMakeSettings.json used with CMake integration for VS2017.

-----

I've been playing with building/testing the library directly in VS2017 IDE.


HDembinski:
Agreed, this is much nicer. I only put the CMakeLists.txt in "build", because I vaguely remember reading that all build scripts should be in "build", but like we discussed privately, I don't remember where I got that from and other Boost libs (hana, compute) also put the CMakeLists.txt in the root directory.

mloskot:
After the fact, the [Steven Watenabe's post](https://lists.boost.org/Archives/boost/2018/09/243361.php) made me realise that it could be even better to refactor as:

```
CMakeListst.txt
test/CMakeListst.txt
examples/CMakeListst.txt
...
```

Let me know if you like this idea, and I may try to prepare a PR.

HDembinski:
The build script is small enough that it is not too confusing to have everyone in one script, so I would not bother changing this for now.

mloskot:
Sure, makes sense.

PR #187, opened 2019-04-28
Add .gitattributes file

mloskot:
Ensure .sh scripts have LF and .bat scripts have CRLF in the working directory.
Avoid conversion of Unix scripts to CRLF when checking out with `core.autocrlf=true`, on Windows, what may cause (Unix) shell failures.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Set default behaviour, in case users don't have core.autocrlf set.
+* text=auto !eol
+
+# Scripts
+*.bat               text eol=crlf
+*.sh                text eol=lf


HDembinski:
So far I didn't have trouble with this, but thank you. I am puzzled because the Travis build didn't run. Could this be a side effect of this patch?

mloskot:
> So far I didn't have trouble with this,

If one clones on Windows and uses the local repo as shared folder in WSL  and VMs, then the bug hits.

>  I am puzzled because the Travis build didn't run. 

Yes, because I told it to skip the builds, with `[skip ci]` in the commit message. 
Since, the patch does not affect builds, I went carbon-friendly  :-)

HDembinski:
Could you please rebase and trigger a full build? I would like to see explicitly that the gitattributes work on travis and appveyor. Then I will merge, thank you!

mloskot:
Rebased.

HDembinski:
Thank you!

mloskot:
Thanks for merging this 'every little helps' :)

PR #190, opened 2019-05-12
Avoid try, catch, throw and use alloc_construct and alloc_destroy

glenfe:
This change:
* Uses RAII instead of try, catch, throw to achieve exception safety with or without exceptions enabled
* Uses boost::alloc_construct and boost::alloc_destroy from Boost.Core to simplify the implementation

Note that you don't need to use allocator_traits for allocate() and deallocate(). Every allocator is required to provide them.

### DIFF
--- a/include/boost/histogram/detail/variant.hpp
+++ b/include/boost/histogram/detail/variant.hpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2019 Hans Dembinski
+// Copyright (c) 2019 Glen Joseph Fernandes (glenjofe@gmail.com)
 //
 // Distributed under the Boost Software License, Version 1.0.
 // (See accompanying file LICENSE_1_0.txt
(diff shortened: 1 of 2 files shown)


glenfe:
The failure appears to be that Appveyor is using Boost.Core from master instead of Boost.Core from develop. I'll merge Core develop to master later today.

glenfe:
> I'll merge Core develop to master later today.

Done.

HDembinski:
Beautiful solution, much better than using macros and makes better use of existing library code. Thank you!

HDembinski:
> On 12. May 2019, at 18:43, Glen Fernandes <notifications@github.com> wrote:
> 
> The failure appears to be that Appveyor is using Boost.Core from master instead of Boost.Core from develop. I'll merge Core develop to master later today.

Yes, I need to fix the Appveyor build so that it correctly uses the develop and master branches, as I am doing already in Travis.

PR #207, opened 2019-08-25
Fix project bug

HDembinski:


HDembinski:
It is difficult to find performance guarantees for std::function, but for boost::function (precursor) they are listed https://www.boost.org/doc/libs/1_71_0/doc/html/function/misc.html

PR #209, opened 2019-09-04
Fix missing include

henryiii:
`make_default` is used in this file, but was not included, causing this to break if `make_default.hpp` was not included before it somewhere.


HDembinski:
Thank you for finding this oversight!

PR #230, opened 2019-10-03
Faster linearize

HDembinski:


HDembinski:
Summary so far:
* introduce offset for faster linearization of non-growing axes
* added traits::is_inclusive and constexpr bool inclusive() methods for builtin axes
* bug fix of fill method when weight array is used with non-inclusive axis
* bug fix of fill method when growing axes are used
* bug fix of axis::options::test(...)
* coverage tested with gcc-8, updated CONTRIBUTING.md with coverage info

Todo:
* document inclusive() method in axis concept

PR #231, opened 2019-10-07
Fixing missing inline, adding link test

henryiii:
There is an inline missing in one of the recent updates, which makes Boost.Histogram unusable when linking two files. I've added a test that fails if this happens again, and fixed the issue.

### DIFF
--- a/test/Jamfile
+++ b/test/Jamfile
@@ -69,6 +69,11 @@ alias cxx17 :
     [ requires cpp_deduction_guides ]
     ;
 
+# Verify that the compile succeeds when linking (missing inline detection) 
+alias compiles :
+    [ link link_2_test.cpp link_1_test.cpp ]
+    ;
+
 # check that useful error messages are produced when library is used incorrectly
 alias failure :
     [ compile-fail axis_category_fail0.cpp ]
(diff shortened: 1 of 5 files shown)


test/Jamfile
```diff
@@ -69,6 +69,11 @@ alias cxx17 :
     [ requires cpp_deduction_guides ]
     ;
 
+# Verify that the compile succeeds when linking (missing inline detection) 
+alias compiles :
+    [ link link_2_test.cpp link_1_test.cpp ]
```
henryiii: Please check my Jamfile, I'm not a regular Jammer. ;)

test/link_1_test.cpp
```diff
@@ -0,0 +1,8 @@
+#include <boost/histogram.hpp>
+#include <boost/histogram/axis/ostream.hpp>
+#include <boost/histogram/ostream.hpp>
```
henryiii: Any other files not included by `<boost/histogram.hpp>` should be added here.

test/link_2_test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+// All files should be included above
+
+int do_nothing();
+
+// Verifies there are no functions with missing inline
+int main() {
+
+  int a = do_nothing();
+  BOOST_TEST_EQ(a, 7);
```
henryiii: I actually only build and link, so this really doesn't even need a BOOST_TEST.

HDembinski:
Thank you, having such a test is a very good idea. There are dedicated link tests in Boost, I will look into this, but for now the test seems to do the job.

PR #232, opened 2019-10-07
Formatting

henryiii:
I've noticed the recent `display.hpp` addition was not properly formatted, so I've fixed that here, along with some .cpp files that were unformatted.

Also added Black formatting to the python files in a separate commit included in this PR.

No extra testing or the like added; should be whitespace and whitespace-like changes only.


HDembinski:
I do not share you view that everything should be formatted in a certain way. I only care about patches being readable.

PR #233, opened 2019-10-09
Improve odr test

HDembinski:


henryiii:
You may need to normalize paths on Windows in the Python script, I see slashes in both directions (`/` and `\`)

test/odr_test.py
```diff
@@ -0,0 +1,48 @@
+import os
+import sys
+import re
+
+this_path = os.path.dirname(__file__)
+
+all_headers = set()
+include_path = this_path + "/../include/"
```
henryiii:
```suggestion
include_path = os.path.join(this_path, "..", "include")
```

test/odr_test.py
```diff
@@ -0,0 +1,48 @@
...
+for root, dirs, files in os.walk(include_path):
+    for fn in files:
+        fn = root + "/" + fn
```
henryiii:
```suggestion
        fn = os.path.join(root,  fn)
```

test/odr_test.py
```diff
@@ -0,0 +1,48 @@
...
+        assert fn.startswith(include_path)
```
henryiii: I would probably make sure these are resolved with `os.path.abspath` first. Might be fine as is though.

test/odr_test.py
```diff
@@ -0,0 +1,48 @@
...
+    with open(filename) as f:
+        for hdr in re.findall('^#include [<"]([^>"]+)[>"]', f.read(), re.MULTILINE):
+            if not hdr.startswith("boost/histogram"):
+                continue
+            yield hdr
+
+
+included_headers = set()
+unread_headers = set()
+for hdr in get_headers(this_path + "/odr_test.cpp"):
```
henryiii:
```suggestion
for hdr in get_headers(os.path.join(this_path, "/odr_test.cpp")):
```

This is why I love pathlib, plumbum.path, pretty much anything OO for paths!

test/odr_test.py
```diff
@@ -0,0 +1,48 @@
...
+    unread_headers.add(hdr)
+
+while unread_headers:
+    included_headers.update(unread_headers)
+    for hdr in tuple(unread_headers):
+        unread_headers.remove(hdr)
+        for hdr2 in get_headers(include_path + hdr):
```
henryiii:
```suggestion
        for hdr2 in get_headers(os.path.join(include_path, hdr)):
```

henryiii:
Instead of path.join everywhere, maybe do a sub of back slash for forward slash, probably on the walk. path.join will actualy put in the backslash, so the strings won't match what you pull out of the source unless you do abspath on them.

test/odr_test.py
henryiii: (These are suggestions to help debug/fix - I have not tested locally and do not have a Windows machine)

henryiii:
I made a PR to your branch with a modified script that *should* work on Windows.

test/odr_test.py
HDembinski: Good point, will do. I just applied that fix to iminuit's setup.py as well the other day.

test/odr_test.py
HDembinski: What difference would abspath make?

test/odr_test.py
HDembinski: I don't mind using os.path.join for the time being.

henryiii:
Have you looked at the version I provided in the PR? It is much simpler, and should have worked out of the box.

HDembinski:
Already merged in 05c1bbfac12531e9f629ae40d1e53f47ffdbf1ee

henryiii:
:+1:

Issue #244, opened 2019-10-26
Indexed iterator does not work on array storage


henryiii:
The following code fails on Wandbox with Boost 1.71 (and similar behavior has been seen in develop):

```cpp
#include <iostream>
#include <cstdlib>
#include <boost/histogram.hpp>

int main()
{
    std::cout << "Hello, Wandbox!" << std::endl;
    
    namespace bh = boost::histogram;
    
    auto h = bh::make_histogram_with(std::array<double, 7>(), bh::axis::integer<>(0,5));
    h(2); h(3); h(3);
    
    for(auto&& ind : bh::indexed(h) )
        std::cout << *ind << " ";
}
```

The failure:

```
In file included from /opt/wandbox/boost-1.71.0/gcc-head/include/boost/histogram/algorithm/project.hpp:16,
                 from /opt/wandbox/boost-1.71.0/gcc-head/include/boost/histogram.hpp:31,
                 from prog.cc:4:
/opt/wandbox/boost-1.71.0/gcc-head/include/boost/histogram/indexed.hpp: In instantiation of 'class boost::histogram::indexed_range<boost::histogram::histogram<std::tuple<boost::histogram::axis::integer<int, boost::use_default, boost::use_default> >, boost::histogram::storage_adaptor<std::array<double, 7> > > >':
/opt/wandbox/boost-1.71.0/gcc-head/include/boost/histogram/indexed.hpp:343:63:   required from 'auto boost::histogram::indexed(Histogram&&, boost::histogram::coverage) [with Histogram = boost::histogram::histogram<std::tuple<boost::histogram::axis::integer<int, boost::use_default, boost::use_default> >, boost::histogram::storage_adaptor<std::array<double, 7> > >&]'
prog.cc:15:35:   required from here
/opt/wandbox/boost-1.71.0/gcc-head/include/boost/histogram/indexed.hpp:47:9: error: 'boost::histogram::indexed_range<boost::histogram::histogram<std::tuple<boost::histogram::axis::integer<int, boost::use_default, boost::use_default> >, boost::histogram::storage_adaptor<std::array<double, 7> > > >::value_iterator' {aka 'double*'} is not a class, struct, or union type
   47 |   using value_reference = typename value_iterator::reference;
      |         ^~~~~~~~~~~~~~~
[...]

PR #245, opened 2019-10-26
Adding empty(hist, cov)

### LINKED ISSUE #241: FEATURE REQUEST: Add empty
I would like an `empty(coverage)` function or method for histograms, something like this:

```cpp
bh::empty(hist, bh::coverage::all);
// OR
hist. empty(bh::coverage::all);
```

It could look something like this:

```cpp
template<class histogram_t>
bool empty(const histogram_t &h, bh::coverage cov) {
    using value_type = typename histogram_t::value_type;
    value_type zero = value_type();
    for(auto&& ind : bh::indexed(h, cov)) {
        if(*ind != zero) {
            return false;
        }
    }
    return true;
}
```

This should be faster than sum if the histogram is filled, and would be useful for checking to see if an item fell into the histogram (inner), or for checking to see if this was filled yet (all).

henryiii:
This closes #241. I did not select a default for coverage, and I was not able to add one test that I wanted to due to bug #244 (not related to this PR, though).

### DIFF
--- a/include/boost/histogram/algorithm/empty.hpp
+++ b/include/boost/histogram/algorithm/empty.hpp
@@ -1,0 +1,36 @@
+// Copyright 2019 Henry Schreiner
+//
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_HISTOGRAM_ALGORITHM_EMPTY_HPP
+#define BOOST_HISTOGRAM_ALGORITHM_EMPTY_HPP
+
+#include <boost/histogram/fwd.hpp>
+#include <boost/histogram/indexed.hpp>
+
+namespace boost {
+namespace histogram {
+namespace algorithm {
+/** Check to see if all histogram cells are empty. Use coverage to include or
+  exclude the underflow/overflow bins.
+
+  This algorithm has O(N) complexity, where N is the number of cells.
+
+  Returns true if all cells are empty, and false otherwise.
+ */
+template <class A, class S>
+auto empty(const histogram<A, S>& h, coverage cov) {
+  using value_type = typename histogram<A, S>::value_type;
+  const value_type default_value = value_type();
+  for (auto&& ind : indexed(h, cov)) {
+    if (*ind != default_value) { return false; }
+  }
+  return true;
+}
+} // namespace algorithm
+} // namespace histogram
+} // namespace boost
+
+#endif
--- a/include/boost/histogram/algorithm.hpp
+++ b/include/boost/histogram/algorithm.hpp
@@ -12,6 +12,7 @@
   Includes all algorithm headers of the Boost.Histogram library.
 */
 
+#include <boost/histogram/algorithm/empty.hpp>
 #include <boost/histogram/algorithm/project.hpp>
 #include <boost/histogram/algorithm/reduce.hpp>
 #include <boost/histogram/algorithm/sum.hpp>
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -62,6 +62,8 @@ boost_test(TYPE run SOURCES algorithm_reduce_test.cpp
   LIBRARIES Boost::histogram Boost::core)
 boost_test(TYPE run SOURCES algorithm_sum_test.cpp
   LIBRARIES Boost::histogram Boost::core)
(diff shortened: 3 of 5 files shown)


include/boost/histogram/algorithm/empty.hpp
```diff
@@ -0,0 +1,38 @@
+// Copyright 2019 Hans Dembinski and Henry Schreiner
```
HDembinski: Since you wrote the code, I think only you should be listed here. I will add my name if I make changes.

include/boost/histogram/algorithm/empty.hpp
```diff
@@ -0,0 +1,38 @@
...
+//
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_HISTOGRAM_ALGORITHM_EMPTY_HPP
+#define BOOST_HISTOGRAM_ALGORITHM_EMPTY_HPP
+
+#include <boost/histogram/coverage.hpp>
```
HDembinski: There is no such header, that's why the tests fail.

include/boost/histogram/algorithm/empty.hpp
henryiii: Okay, I didn't know what the right thing to put here was.

include/boost/histogram/algorithm/empty.hpp
henryiii: Oops, missed that. I moved it to a new header, then realized it wasn't required for this PR since I'm using `indexed`, so left that off the PR. Forget I imported it here. My computer still has the header so it doesn't fail locally.

HDembinski: Since the coverage enum is used by several algorithms now, it should indeed be moved into another header, but the header should be boost/histogram/algorithm/coverage.hpp

HDembinski: Actually, no, it should just live in fwd.hpp.

include/boost/histogram/algorithm/empty.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+  All cells in the histogram are checked, and this returns as soon as one cell is
+  found that is not empty.
+
+  Returns true if all cells are empty, and false otherwise.
+ */
+template <class A, class S>
+auto empty(const histogram<A, S>& h, coverage cov) {
+  using value_type = typename histogram<A, S>::value_type;
+  value_type zero = value_type();
```
HDembinski: This can be const and I think it should be "default_value" instead of "zero".

include/boost/histogram/algorithm/empty.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <boost/histogram/indexed.hpp>
+
+namespace boost {
+namespace histogram {
+namespace algorithm {
+/** Check to see if all histogram cells are empty. Use coverage to include or
+  exclude the underflow/overflow bins.
+
+  All cells in the histogram are checked, and this returns as soon as one cell is
+  found that is not empty.
```
HDembinski: Better replace by: This algorithm has O(N) complexity, where N is the number of cells.

HDembinski:
Otherwise looks very good. Almost ready to merge!

include/boost/histogram/algorithm/empty.hpp
henryiii: :+1:

include/boost/histogram/algorithm/empty.hpp
henryiii: It seems odd to have something live in fwd that doesn't live anywhere else...

include/boost/histogram/algorithm/empty.hpp
henryiii: Yes.

include/boost/histogram/algorithm/empty.hpp
henryiii: Though, since it is always part of a function call, and can't be used on its own, I can see why it could make sense there.

henryiii:
I think I see my mistake with the test. I can fix when I get home.

henryiii:
I'm quite surprised clang allowed `unsigned i = -1`... Anyway, that hid a logic error in test 2. I think we should pass tests now, fingers crossed.

HDembinski:
Thanks!

Issue #250, opened 2019-11-11
use arguments of storage to generate accumulator default value


HDembinski:
The storage is passed as a value to the histogram, because it allows the storage to be configured at run-time.
```
auto h = make_histogram_with(dense_storage<int>(), ...);
```

So far we have not made use of this feature, but I propose to extend the storage adaptor so that it accepts arguments in its ctor which are used to initialize the default value for its accumulator.
```
auto h = make_histogram_with(dense_storage<int>(3), ...);
```
Internally, the storage creates a value `int default_value{3}` with the arguments passed to it. Then, whenever the histogram resets the storage, the accumulators are initialized to this value.

This feature allows one to pass run-time configuration to the accumulators, which is currently not possible.

@henryiii @jpivarski any thoughts?

henryiii:
I can't come up with an obvious use case for this, so can't be much help here. This would fill flow bins with the same default, and uneven binning scheme with a single default too. A function that takes an indexed-like argument would be more useful (but rather heavyweight for this).

The thing I'd like to see is a way to take an existing piece of memory (ideally supporting numpy-like strides and offsets) in and use that, ideally without a copy. This would be useful in converting numpy arrays and/or aghast flatbuffers (I think) into histograms without making a copy. In C++, this could be used to take eigen/XTensor (which I think have the same memory scheme), as well as could possibly be used to take over a ROOT histogram without a copy (probably not a weighted one, however).

HDembinski:
It is open what the accumulator does with the initial arguments. For the currently builtin accumulators, this feature makes not much sense. But you could come up with some use-defined accumulator that needs some run-time configuration. Right now, you cannot pass this information to the accumulator. This feature is supposed to change this.

Here is a use case. You want to store multiple weights, but you only know how many you need at run-time.

```
struct dynamic_weighted_sum {
  dynamic_weighted_sum(unsigned nweights) : weights_(nweights, 0);
  // ...
  std::vector<double> weights_;
}
```

jbuonagurio:
I came across this in the process of creating an accumulator for rolling means with `window_size` defined at run-time. Basic implementation follows, which should handle unsigned value types.

I could try to put together a PR. Any thoughts on the best way to extend `storage_adaptor` without breaking the forwarding constructor to underlying containers? Possibly create a tag type specifically for accumulator arguments?

```cpp
template <class ValueType>
struct rolling_mean
{
    using value_type = ValueType;
    using const_reference = const value_type&;
  
[...]

PR #269, opened 2020-02-29
implement rolling mean accumulator

jbuonagurio:
As requested in #250, this PR includes a basic rolling mean accumulator for trivial value types and a basic circular buffer implementation to eliminate the dependency on `boost::circular_buffer`.

Work-in-progress while awaiting support for arguments in `storage_adaptor`. Will of course help with unit tests as well!

### DIFF
--- a/include/boost/histogram/accumulators/rolling_mean.hpp
+++ b/include/boost/histogram/accumulators/rolling_mean.hpp
@@ -1,0 +1,178 @@
+#include <boost/assert.hpp>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace boost {
+namespace histogram {
+namespace accumulators {
+namespace detail {
+
+template <class T, class Allocator = std::allocator<T>>
+struct circular_buffer
+{
+  using value_type = typename std::allocator_traits<Allocator>::value_type;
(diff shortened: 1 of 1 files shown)


jbuonagurio:
I added the simple forward iterator to `circular_buffer` just for debugging purposes.

HDembinski:
@jbuonagurio Do you agree with closing this?

jbuonagurio:
> @jbuonagurio Do you agree with closing this?

@HDembinski that's up to you, this PR was per your request in #250. They are important to me, but are rolling calculations important enough to other users to warrant an implementation at the library-level? I could continue work on this, adding support for variance calculation and compile-time fixed-size buffers. The circular buffer implementation is close to as simple as it gets unless I use something like `std::deque` which would probably impact performance, or depend on `boost::circular_buffer` (as Boost.Accumulators does).

HDembinski:
I thought it is a good idea at first, but now it looks like we are rebuilding the rolling mean from Boost.Accumulators, which does not sound like a good investment of our time. The library-included accumulators mean, weighted_mean, etc are also duplications of code in Boost.Accumulatos, but they are so common and so simple that they can be implemented without relying on Boost.Accumulators explicitly. However, this should be rather the exception than the rule. Code duplication is bad.

PR #271, opened 2020-03-05
[Fix] add language feature-macro for <span> for detail::span

### LINKED ISSUE #270: Problem compiling with clang-9 and libc++ and c++17
Hi,

I am using this great library in a project that compiles with C++17 but up to date toolchains.

I believe the condition at
https://github.com/boostorg/histogram/blob/a872c6e1c32e3c6000317b07aec7e416c8939c75/include/boost/histogram/detail/span.hpp#L11
is not fully correct, because e.g. clang-9 fullfills all those conditions, and has the include for `<span>`, but libc++ does not enable this include for C++17.

Adding a `__cpp_lib_span >= 201902` as condition does fix this issue. (see https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B20)

Thanks again for the great library!

Best Regards, Jonas

JonasToth:
This commit fixes an issue with newer clang/libc++-versions supporting C++2a.
When compiling in C++17 (or earlier) mode, the compiler and libc++ do not expose
the `<span>` include. The logic for `detail::span` to include `<span>`
instead of providing an own implementation does only check if the
include `<span>` exists. It exists, but it is not exposed.
Therefore the compilation fails without this fix complaining about the missing `span`.

Fixes #270

### DIFF
--- a/include/boost/histogram/detail/span.hpp
+++ b/include/boost/histogram/detail/span.hpp
@@ -8,7 +8,8 @@
 #define BOOST_HISTOGRAM_DETAIL_SPAN_HPP
 
 #if __cpp_constexpr >= 201603 && __cpp_deduction_guides >= 201703 && \
-    __cpp_lib_nonmember_container_access >= 201411 && __has_include(<span>)
+    __cpp_lib_nonmember_container_access >= 201411 && \
+    __cpp_lib_span >= 201902 && __has_include(<span>)
 #include <span>
 
 namespace boost {


henryiii:
This macro is supposed to be defined *in* `<span>`, so the check needs to be a little more complex. It should happen after the include.

HDembinski:
Thank you very much. The failing test is unrelated to this patch, I will look into it tonight and patch this.

HDembinski:
> This macro is supposed to be defined _in_ `<span>`, so the check needs to be a little more complex. It should happen after the include.

According to the docs,
https://en.cppreference.com/w/cpp/utility/feature_test
we need to include either <version> or <span> to have the feature test macro, <version> seems to be more straight forward, but also has to be protected with a `__has_include`. So yes, this test needs to be more complex.

include/boost/histogram/detail/span.hpp
```diff
@@ -8,7 +8,8 @@
 #define BOOST_HISTOGRAM_DETAIL_SPAN_HPP
```
henryiii:
```suggestion
#if __has_include(<version>)
#include <version>
#endif
```

Something like this, perhaps?

HDembinski: Yes

HDembinski:
@JonasToth It turns out I already had a fix for this in my private fork, which I had forgotten to upload to the main repository. I updated the develop branch now which has the fix for this issue.

JonasToth:
Perfect, thank you! I didnt had time to recheck yet, but as its fixed now, even better :)

Do you know, when the next boost-release will be?

HDembinski:
Next release is in April
https://github.com/boostorg/wiki/wiki/Releases%3A-Schedule

PR #283, opened 2020-07-03
fix: Boolean axis missing options() method

henryiii:
The new boolean axis is missing an `options()` method, which causes it to break in code that expects all axes to have an options method (boost-histogram, for example). This adds this method (hopefully in time for 1.74).

### DIFF
--- a/include/boost/histogram/axis/boolean.hpp
+++ b/include/boost/histogram/axis/boolean.hpp
@@ -9,6 +9,7 @@
 
 #include <boost/core/nvp.hpp>
 #include <boost/histogram/axis/iterator.hpp>
+#include <boost/histogram/axis/option.hpp>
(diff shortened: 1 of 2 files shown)


HDembinski:
Thank you, I will merge this, but your code should not rely on options(), it is an optional axis feature, see https://www.boost.org/doc/libs/develop/libs/histogram/doc/html/histogram/concepts.html

Please use axis::traits::options() for uniform access.

henryiii:
Thank you, I think this is still more consistent; Boost.Histogram should provide the optional features if possible -  It's great that it's not required, as well, though. 

Actually, for my usage, `axis::traits::get_options` seems to be perfect, adapted the boost-histogram code to consistently use that. :)

PR #301, opened 2020-11-10
bug-fix: handling of special values in ctor and in value method

HDembinski:
Fixed
* Passing nan to the ctor of axis::variable failed to throw std::invalid_argument
* axis::variable{{0, 1, inf}}.value(1) was nan, because interpolation would compute 0 * inf


HDembinski:
Affected by bug and fix: boost-histogram, related issue https://github.com/scikit-hep/boost-histogram/issues/468

HDembinski:
Closes scikit-hep/boost-histogram#468

PR #315, opened 2021-03-18
Support floating point numbers in thread_safe accumulator

HDembinski:
This adds a workaround for floating point numbers that I learned from discussion with @emanca.

We further stop inheriting from `std::atomic` for better encapsulation.


HDembinski:
@henryiii @jpivarski Any comments on this?

Not inheriting from std::atomic anymore is technically interface-breaking, but people should not use this accumulators directly, so I think it is acceptable.

PR #324, opened 2021-04-28
Update lcov coverage

sdarwin:
Hi,

Some info about the changes:
- upgrade lcov from 1.14 to 1.15
- adds support for lcov with gcc-9 and gcc-10.  However, lcov 1.15 processing gcc-9 and 10 is rather slow.
- gcov ideally shouldn't be hardcoded to version 8. Rather, match the gcc version. Thus, GCOVTOOL='gcov-4.6' for gcc-4.6, or GCOVTOOL='gcov-6' for gcc-6.
- btw, the new cov.yml file was based on https://github.com/boostorg/boost-ci/blob/master/ci.yml, which is a generic GHA file that pdimov wrote to use with any boost library.

> I also want to generate coverage data offline.

That should be fairly simple, by running the same CI scripts on a local machine.   I will test it out, and send over the steps.

### DIFF
--- a/.github/workflows/cov.yml
+++ b/.github/workflows/cov.yml
@@ -10,54 +10,129 @@ on:
       - 'examples/**'
       - 'tools/**'
       - '*.md'
+  push:
+    branches:
+      - master
+      - develop
+      - feature/**
+      - test*
 
 env:
+  UBSAN_OPTIONS: print_stacktrace=1
   B2_OPTS: -q -j2 warnings-as-errors=on
 
 jobs:
-  cov:
-    runs-on: macos-10.15
+  posix:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - toolset: gcc-5
+            gcov: gcov-5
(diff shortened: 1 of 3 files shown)


sdarwin:
The most recent commit includes "tools/cov_offline.sh".    Running this standalone script will - 
- checkout the repo (if it hasn't already been done) 
- build boost
- build the library
- run lcov, generating coverage.info
- optionally upload the file to coveralls.io

HDembinski:
Cool, thanks, sorry for the late reply.

> gcov ideally shouldn't be hardcoded to version 8. Rather, match the gcc version. Thus, GCOVTOOL='gcov-4.6' for gcc-4.6, or GCOVTOOL='gcov-6' for gcc-6.

Sure, the problem is that no other version of gcc worked.

HDembinski:
After careful reading I am going to reject this. This is a very invasive patch that changes at lot of things that do not need changing.

The salient change seems to be to upgrade to lcov-1.15.

PR #326, opened 2021-05-16
deprecate accumulators::thread_safe, replace with optionally thread-safe accumulators::count

HDembinski:
The implementation of thread_safe is reverted to that of the last release v1.76


HDembinski:
@henryiii FYI

PR #343, opened 2021-09-30
deprecate sum::small and sum::large as workaround for bug in windows.h

### LINKED ISSUE #342: windows.h illegally uses `#define small char`, explore possible work-arounds
According to my research, it is not possible to work around this error in the windows.h header in this library without breaking the API.

https://stackoverflow.com/questions/21165891/is-small-a-keyword-in-c

We could deprecate the old API for everyone and add new methods `large_part` and `small_part` as suggested in #341. When the preprocessor detects that `small` is set, it would not compile the deprecated API. This should fix the issue for those that are affected but not break the code of other people.

HDembinski:
Closes #342 

This is based on the discussion in #341, but without immediately breaking the API. Instead, the old API is deprecated in favor of an API that works around the issue.

We should not have to work around Microsoft violating the C++ standard, but here we are.


HDembinski:
@emmenlau FYI

emmenlau:
Oh my, I'm so sorry it had to come to this! But its highly appreciated and will possibly help many users in the long run. Thanks a lot @HDembinski for all your effort!

henryiii:
Good workaround for an unfortunate issue. In boost-histogram, should we move to using the new names or keep the old names (since they are just strings in pybind11, and Python obviously supports large and small). The old names are shorter and easier to remember/use in the REPL, etc, but the longer names match Boost.Histogram better. (and the sum accumulator doesn't have much use except for teaching about accumulators or for doing a manual highly-accurate accumulation, where most users won't need the two parts). We could even provide the official names, but have the short names as non-deprecated shortcuts if you want.

Issue #307, opened 2021-03-08
Axis with closed bounds in the last bin


jolopezl:
(See: https://gitter.im/boostorg/histogram?at=6040fcdc457d6b4a94a1e3be)

I wonder if there is a proper way to set the last bin of a histogram to have a closed bound? It has some statistical effects and applications when analyzing images. Maybe we can foresee a possible way to set a switch for this?

The axis would look like this: `[x_0),x_1),...),x_i),...,x_N-1),x_N]`, instead of `[x_0),x_1),...),x_i),...,x_N-1),x_N)`.

henryiii:
This is pretty easy to implement if you need it, see https://github.com/scikit-hep/boost-histogram/blob/6170cff72d65e63c1f4ad0504b3dab16152cec6b/include/bh_python/regular_numpy.hpp .

PR #344, opened 2021-09-30
regular and variable axis are closed intervals if overflow bin is absent

### LINKED ISSUE #319: Proposal of API change: regular axis without overflow bin is inclusive on upper edge of last bin
The proposal is to make the upper edge of the last bin of a regular axis inclusive, but only if overflow is off. One could argue that this is closer to what the user expects. It would address #307 in a natural way and would make the NumpyAxis in boost-histogram obsolete. It would also mean that we do not have to introduce a special axis for probabilities, where you do not need overflow bins, but you can certainly have probabilities equal to 1, which you want to include in the last bin.

This is - strictly speaking - a breaking change since code that used such a regular axis would produce different results once this is changed, but it is a corner case. If you used a regular axis without overflow and you have values that end up exactly on the upper edge of the last bin, would you not want them to be included? I cannot imagine a scenario where the intent is to exclude them. So this could be regarded as a "fix" rather than a breaking change.

@henryiii @jpivarski ?

### LINKED ISSUE #336: How to bin integral vs floating point correctly (precision issue and general question)
Dear @HDembinski and other authors, thanks for this awesome library. We have just started using it and the speed and versatility are impressive! We have a problem that manifests as a precision problem, but also shows a more general question.

The problem: We use the floating precision histogram to bin any kind of data. When we bin data (gray value camera images) of integer precision, we set lower and upper bound based on the data range to `[min, max+1]` and the number of bins to `upper bound - lower bound`. However there are sometimes cases where the numeric precision then leads to an empty last bin. This is due to rounding error, the values fall into `max-1` instead of `max`. We checked and the bin index is then computed as `(max-1).999...` which finally resolves to `max-1`. Is there a good solution to this? It leads to confusion that the last bin is empty, and its also a bit problematic in writing tests.

The underlying question: We are not completely certain what is a "recommended" way to set histogram bounds for integral data space. Are there common guidelines? In floating point precision it is more understandable what data range is covered by each bin. But in integral space, it seems the value `0` could be best represented by bin `[-0.5, 0.5)` rather than the more obvious `[0, 1)`. And help, links or pointers to documentation would be highly appreciated!

HDembinski:
Closes #319, #307
Potentially closes #336


HDembinski:
@emmenlau could you please check whether this patch fixes the issue that you reported in #336

emmenlau:
Sorry for the delay in feedback. This PR is appreciated, we have a use case where this is relevant for us! Thanks.

PR #346, opened 2021-11-05
Update metadata

sdarwin:
In the metadata, the name field is generally being used verbatim to display the library name, even including capitalization.   Compare to other boost libraries...

### DIFF
--- a/meta/libraries.json
+++ b/meta/libraries.json
@@ -1,6 +1,6 @@
 {
     "key": "histogram",
-    "name": "histogram",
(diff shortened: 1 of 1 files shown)


henryiii:
macos-latest now points at macOS 11; I think that’s why this is failing. There’s likely a bit removed from the image; a brew install gsl or something is likely needed. On vacation so can’t check further right now. If @HDembinski can’t get to it this weekend I can help early next week.

PR #361, opened 2022-07-23
Add fraction accumulator (for calculating efficiencies, etc)

gohil-jay:
Added standard efficiency accumulator with value(), variance() and confidence_interval()

Working [here](https://godbolt.org/z/4dqbWh5jr).


HDembinski (changes requested):

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+// Distributed under the Boost Software License, version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_HISTOGRAM_ACCUMULATORS_EFFICIENCY_HPP
+#define BOOST_HISTOGRAM_ACCUMULATORS_EFFICIENCY_HPP
+
+#include <boost/core/nvp.hpp>
+#include <boost/histogram/detail/atomic_number.hpp>
+#include <boost/histogram/fwd.hpp> // for efficiency<>
```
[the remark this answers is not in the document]
gohil-jay: Fixed now.

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <boost/format.hpp>
+#include <boost/histogram.hpp>
+#include <cmath>
+
+enum interval_type {
+    Wald=0,
+    Wilson=1,
+    Jeffreys=2,
+    ClopperPearson=3,
+    AgrestiCoull=4
```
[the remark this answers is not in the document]
gohil-jay: Done :)

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+struct efficiency {
+    // return value is ignored, so we use void
+    void operator()(bool x) {
+        if (x) ++n_success_;
+        else ++n_failure_;
+    }
+
+    double value() const { return n_success_/(n_success_+n_failure_); }
+    
+    double variance() const { return n_success_*n_failure_*(n_success_+n_failure_); }
```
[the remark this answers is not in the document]
gohil-jay: Added source Wikipedia in comment

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+    double confidence_interval(interval_type CI_type = Wald, int conf_level = 95) const {
```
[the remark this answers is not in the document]
gohil-jay: All variables changed.

Changed to std::pair<double, double> too (godbolt [here](https://godbolt.org/z/sx4Edj41Y))

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+        else if (conf_level >= 95 && conf_level < 98) {z = 1.96; }
+        else {z = 1.645; }
+        
+        switch(CI_type)
+        {
+            case Wald: return (z*pow((n_failure_*n_success_), 0.5)) / pow(n_failure_+n_success_, 1.5);
+            case Wilson: return (z/(n_failure_+n_success_+pow(z, 2)))*pow(((n_failure_*n_success_)/(n_failure_ + n_success_))+(pow(z,2)/4), 0.5);
+            case Jeffreys: return 0; // implement if needed
+            case ClopperPearson: return 0; // implement if needed
+            case AgrestiCoull: return 0; // implement if needed
```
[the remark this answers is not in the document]
gohil-jay: All done

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+    double value() const { return n_success_/(n_success_+n_failure_); }
+    
+    double variance() const { return n_success_*n_failure_*(n_success_+n_failure_); }
+
+    double confidence_interval(interval_type CI_type = Wald, int conf_level = 95) const {
+        double z;
+        if (conf_level >= 99) {z = 2.576; }
+        else if (conf_level >= 98 && conf_level < 99) {z = 2.326; }
+        else if (conf_level >= 95 && conf_level < 98) {z = 1.96; }
+        else {z = 1.645; }
```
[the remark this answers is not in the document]
gohil-jay: This was most fun, but done!

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,59 @@
...
+        
+        switch(CI_type)
+        {
+            case Wald: return (z*pow((n_failure_*n_success_), 0.5)) / pow(n_failure_+n_success_, 1.5);
+            case Wilson: return (z/(n_failure_+n_success_+pow(z, 2)))*pow(((n_failure_*n_success_)/(n_failure_ + n_success_))+(pow(z,2)/4), 0.5);
+            case Jeffreys: return 0; // implement if needed
+            case ClopperPearson: return 0; // implement if needed
+            case AgrestiCoull: return 0; // implement if needed
+        };
+        return 0;
```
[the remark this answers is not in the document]
gohil-jay: Returned std::pair with both values `std::numeric_limits<double>::quiet_NaN()` !

include/boost/histogram/detail/wilsonInterval.hpp
```diff
@@ -0,0 +1,25 @@
...
+#define BOOST_HISTOGRAM_DETAIL_WILSON_INTERVAL_HPP
+
+#include <cmath>
+
+namespace boost {
+namespace histogram {
+namespace detail {
+
+template <class T>
+T wilsonInterval(T n_failure_, T n_success_, T z) {
```
HDembinski, line 17: looks good, but use snake_case: `wilson_interval`

include/boost/histogram/detail/wilsonInterval.hpp
```diff
@@ -0,0 +1,25 @@
...
+  return (z/(n_failure_+n_success_+(std::pow(z, 2))))*(std::pow(((n_failure_*n_success_)/(n_failure_ + n_success_))+((std::pow(z,2))/4), 0.5)); // Source: Wilson Interval from Confidence Interval | Wikipedia
```
HDembinski, line 18: Add a comment where you got that formula from (link to Wikipedia)

include/boost/histogram/detail/waldInterval.hpp
```diff
@@ -0,0 +1,25 @@
...
+#define BOOST_HISTOGRAM_DETAIL_WALD_INTERVAL_HPP
+
+#include <cmath>
+
+namespace boost {
+namespace histogram {
+namespace detail {
+
+template <class T>
+T waldInterval(T n_failure_, T n_success_, T z) {
```
HDembinski, line 17: snake_case

include/boost/histogram/detail/waldInterval.hpp
```diff
@@ -0,0 +1,25 @@
...
+  return (z*(std::pow((n_failure_*n_success_), 0.5))) / std::pow(n_failure_+n_success_, 1.5); // Source: Wald Interval from Confidence Interval | Wikipedia
```
HDembinski, line 18: comment where this formula comes from

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <boost/histogram/detail/waldInterval.hpp>
+#include <boost/histogram/detail/wilsonInterval.hpp>
+#include <utility>
+#include <limits>
+
+enum class interval_type {
+    wald,
+    wilson,
+    jeffreys,
+    agrestiCoull,
```
HDembinski, line 22: agresti_coull

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,54 @@
...
+    clopperPearson
```
HDembinski, line 23: clopper_pearson

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+   tt1 = 2/(PI*0.15449436008930206298828125) + 0.5f * lnx;
+   tt2 = 1/(0.15449436008930206298828125) * lnx;
+
+   return(sgn*sqrtf(-tt1 + sqrtf(tt1*tt1 - tt2)));
+}
+
+template <class T>
+T probit(T x){
+   return (-1)*((std::pow(2,0.5))*ErfInv((2*((1-x)/2))-1)) ;
```
HDembinski, line 34: std::sqrt(2) is faster and more readable than std::pow(2, 0.5). Instead of dividing by 2 it is faster to multiply with 0.5. Avoid using integers, all literals in a templated function like this should be encapsulated with a T, for example, 1 becomes T(1) and 0.5 becomes T(0.5).

include/boost/histogram/detail/waldInterval.hpp
HDembinski: T(0.5) instead of 0.5 etc.

include/boost/histogram/detail/waldInterval.hpp
HDembinski: We don't use the trailing underscore for parameters, only for private members of a class.

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <cmath>
+
+# define PI 3.14159265358979323846  /* pi */
+
+namespace boost {
+namespace histogram {
+namespace detail {
+
+template <class T>
+T ErfInv(T x){
```
HDembinski, line 19: snake_case

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,41 @@
...
+   T tt1, tt2, lnx, sgn;
+   sgn = (x < 0) ? -1.0f : 1.0f;
+
+   x = (1 - x)*(1 + x);        // x = 1 - x*x;
+   lnx = logf(x);
```
HDembinski, line 24: std:log instead of logf

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+   tt1 = 2/(PI*0.15449436008930206298828125) + 0.5f * lnx;
```
HDembinski, line 26: `PI` is not available on all platforms. The best is to replace PI*0.15449436008930206298828125 by 0.4853583466776075

HDembinski: Or better just replace 2/(...) by 4.120666747961526

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,41 @@
...
+# define PI 3.14159265358979323846  /* pi */
+
+namespace boost {
+namespace histogram {
+namespace detail {
+
+template <class T>
+T ErfInv(T x){
+   T tt1, tt2, lnx, sgn;
+   sgn = (x < 0) ? -1.0f : 1.0f;
```
HDembinski, line 21: In C++, it is best to use std::copysign, https://en.cppreference.com/w/cpp/numeric/math/copysign

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,41 @@
...
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt
+// or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_HISTOGRAM_DETAIL_PROBIT_HPP
+#define BOOST_HISTOGRAM_DETAIL_PROBIT_HPP
+
+#include <cmath>
+
+# define PI 3.14159265358979323846  /* pi */
```
HDembinski, line 12: We only use the preprocessor as last resort, because it is evil. Best is to just replace PI in the code below with a literal, see comment there. Please remove line 12.

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,54 @@
...
+    std::pair<double, double> confidence_interval(interval_type type = interval_type::wald, double cl = 0.683) const {
+        double z = detail::probit(cl);
+        double p = n_success_/(n_success_+n_failure_);
+        switch(type)
+        {
+            case interval_type::wald: return std::make_pair((p-(detail::waldInterval(n_failure_, n_success_, z))), (p+(detail::waldInterval(n_failure_, n_success_, z))));
+            case interval_type::wilson: return std::make_pair((p-(detail::wilsonInterval(n_failure_, n_success_, z))), (p+(detail::wilsonInterval(n_failure_, n_success_, z))));
+            case interval_type::jeffreys: return std::make_pair(0, 0); // implement if needed
+            case interval_type::clopperPearson: return std::make_pair(0, 0); // implement if needed
+            case interval_type::agrestiCoull: return std::make_pair(0, 0); // implement if needed
```
HDembinski, line 45: I think it is easier and more readable if detail::waldInterval etc returns std::pair, which saves you some repetition here and some inefficiency: you are calling the function twice with the same arguments.

HDembinski: In other words, `detail::waldInterval` and the others should accepted `p`, `n_failure_`, `n_success_` and `z`.

include/boost/histogram/accumulators/efficiency.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+struct efficiency {
+    void operator()(bool x) {
+        if (x) ++n_success_;
+        else ++n_failure_;
+    }
+
+    double value() const { return n_success_/(n_success_+n_failure_); }
+    
+    double variance() const { return n_success_*n_failure_*(n_success_+n_failure_); } // Source: Variance from Binomial Distribution | Wikipedia
```
HDembinski, line 34: line too long, put the comment inside the function
```
double variance() const {
   // Source ...
   return n_success_ ...
}

File: include/boost/histogram/accumulators/efficiency.hpp [hunk 22]
```diff
@@ -0,0 +1,54 @@
...
+
+enum class interval_type {
+    wald,
+    wilson,
+    jeffreys,
+    agrestiCoull,
+    clopperPearson
+};
+
+struct efficiency {
```
HDembinski, line 26: You describe efficiency in fwd.hpp as a templated type, but this type is not templated. We eventually want to make efficiency templated on the type of the counters, right now it is hard-coded to double. We want to eventually make efficiency templated, but I think you should first write some unit tests and then template this class once the unit tests are in place.

HDembinski:
Hi @gohil-jay , great progress! I added new comments. I think the next step for you is to write some unit tests. The tests are located in the folder tests. We do the testing with the lightweight tests in Boost Core. Basic documentation is here: https://www.boost.org/doc/libs/1_79_0/libs/core/doc/html/core/lightweight_test.html.

Have a look how testing is done for other accumulators and try to make a minimal test for your efficiency accumulator. You could also start simpler and test your `detail::waldInterval` (that you should rename to detail::wald_interval`.

HDembinski:
Most of the tests are failing, because I made a mistake, I am fixing that now.

HDembinski:
Please merge the develop branch into this one to get the fix that I just commited. In the terminal, do `git merge develop` when the currently active checkout is your branch.

gohil-jay:
This is what I see...

image

HDembinski:
I didn't find a nice way of fixing the warning, so you have to use the ugly way. Something along these lines will silence the warning:

```cpp
#include <array>
#include <boost/config/workaround.hpp>

int main() {

#if BOOST_WORKAROUND(BOOST_MSVC, >= 0)
#pragma warning(disable : 4305)
#pragma warning(disable : 4838)
#endif
  std::array<float, 2> x = {1.5, 2.1};
#if BOOST_WORKAROUND(BOOST_MSVC, >= 0)
#pragma warning(default : 4305)
#pragma warning(default : 4838)
#endif

  return 0;
}
```

HDembinski:
Can you please do `git checkout develop & git pull & git merge develop` again? This will fix the issue you had with the DEBUG macro on Windows.

gohil-jay:
Done!

HDembinski:
Hi Jay, looks like all tests pass now, but you still need to fix the warning in detail_probit_test with the code in my earlier comment.

I have been thinking about the interface of this class, and we need to iterate it some more. Here are the requirements.

- The class should have minimal size, because we are potentially creating a lot of them. Minimal is two counters for successes and failures.
- We do not want to compute the conversion from CL to z every time someone is calling `confidence_interval`, because this is going to be called a lot and the conversion is expensive. We could pre-compute this when the class is created, but this is still expensive, since the code will be run for each accumulator instance that is created, and we potentially create a lot of them. Also, we then have to store the result of the computation in the class, and then the class is not of minimal size.
- Users are not going to switch the CL or the type of confidence interval dynamically, they will settle for one setting and then keep going with that. This means we can make these configurations part of the type, by using template parameters. Floating point numbers cannot be used as non-type template parameters, but we can use `std::ratio` instead https://stackoverflow.com/questions/2183087/why-cant-i-use-float-value-as-a-template-parameter.
- We should allow users to pass `z` directly without going through the CL to `z` conversion, because at least in physics we often express confidence levels in `z` sigma  (number of standard deviations of a normal distribution). 5 sigma is easier to mentally parse than 0.9999994266968563.

I will play around with a few ideas on Godbolt. Once I have something, I will present it here. You should then let me know when you are done with your current changes so that I can integrate this part without us creating conflicts for each other by making changes simultaneously.

HDembinski:
@gohil-jay After looking at the Clopper-Pearson interval, I think we cannot avoid to use Boost Math at least for this interval. I don't want to re-implement the regularised incomplete beta function in our `detail` namespace.

include/boost/histogram/detail/probit.hpp
```diff
@@ -0,0 +1,37 @@
...
+template <class T>
+T erf_inv_approx(T x){
+   T one = T(1);
+   T lnx = std::log((one - x)*(one + x));
+
+   T tt1 = T(4.120666747961526) + T(0.5) * lnx;
+   T tt2 = T(6.47272819164) * lnx;
+
+   return std::copysign(std::sqrt(-tt1 + std::sqrt(tt1*tt1 - tt2)), x);
+}
```
HDembinski, line 25: I think it is sufficient to have erf_inv_approx in double precision, and I also think that the `approx` should be in front, so please replace this with this slightly more optimized version
```cpp
    inline double approx_erf_inv(double x) {
        const double lnx = std::log((1 - x)*(1 + x));
        const double tt1 = std::fma(0.5, lnx, 4.120666747961526);
        const double tt2 = 6.47272819164 * lnx;
        return std::copysign(std::sqrt(-tt1 + std::sqrt(std::fma(tt1, tt1, -tt2))), x);
    }
```
`inline` is needed when you define a function inside a header and it is not templated.

HDembinski: std::fma does the operation a * x + b in one CPU cycle. We need to use this explicitly here, because the compiler cannot optimize a * x + b into the fma instruction automatically unless special compiler flags are used (which are usually not turned on). The optimizer is not allowed to perform code transformations that could produce a different result.

HDembinski: You can ignore this request, because I made the change in my commit to your PR. I replaced detail/probit.hpp with detail/normal.hpp and improved the erf_inv implementation. I still leave the comments here for you to read.

HDembinski (approved):

HDembinski:
@gohil-jay We need do some more work on this branch, but the "changes requested" warning is annoying me, so I accepted the changes you made so far.

HDembinski:
I added two new classes which allow us to define the interval via confidence level or via a multiple of the standard deviation. In physics, this is common can now be easily supported.

HDembinski:
I committed another large patch. The `efficiency` accumulator is now called `fraction`, since this is a more general name.

The fraction accumulator only offers the default interval, which is the Wilson score interval. Users can compute other intervals with the interval functors in the utility namespace which are publically exposed. I implemented the wilson_interval as an example for you, @gohil-jay , could you please implement the others? You will need special functions from Boost Math for clopper_pearson_interval and jeffreys_interval.

include/boost/histogram/utility/clopper_pearson_interval.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+template <class ValueType>
+class clopper_pearson_interval : public binomial_proportion_interval<ValueType> {
+  using base_t = binomial_proportion_interval<ValueType>;
+
+public:
+  using value_type = typename base_t::value_type;
+  using interval_type = typename base_t::interval_type;
+
+  explicit clopper_pearson_interval(deviation d = deviation{1.0}) noexcept
```
HDembinski, line 26: You should accept a `confidence_level` here instead of a `deviation`. Since `deviation` and `confidence_level` are programmed to be convertible into each other, you can chose where which suits you best / requires less conversion in your implementation below in `operator()`. The Clopper-Pearson interval naturally needs the CL to compute alpha.

include/boost/histogram/utility/clopper_pearson_interval.hpp
```diff
@@ -0,0 +1,58 @@
...
+      : z_{static_cast<double>(d)} {}
+
+  interval_type operator()(value_type successes, value_type failures) const noexcept {
+    const value_type half{0.5}, one{1.0};
+    const value_type ns = successes, nf = failures;
+    const value_type n = ns + nf;
+    // Source: https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval
+    const value_type alpha1 = (one - d) * half;
+    const value_type alpha2 = one - ((one - d) * half);
```
HDembinski, line 35: This part looks wrong, there is only one alpha.

HDembinski: And you cannot compute alpha from d, also there is no variable d in this scope here?

include/boost/histogram/utility/clopper_pearson_interval.hpp
```diff
@@ -0,0 +1,58 @@
...
+    // Source: https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Clopper%E2%80%93Pearson_interval
+    const value_type m1 = ns;
+    const value_type n1 = n - ns + one;
+    const value_type m2 = ns + 1;
+    const value_type n2 = n - ns;
+    // Source: https://en.wikipedia.org/wiki/Beta_distribution
+    const value_type beta1 = (tgamma(m1)*tgamma(n1)) / tgamma(m1+n1);
+    const value_type beta2 = (tgamma(m2)*tgamma(n2)) / tgamma(m2+n2);
```
HDembinski, line 43: I am not sure what you are calculating here. You need the quantiles of the beta distribution to compute the interval.

https://en.wikipedia.org/wiki/Beta_distribution

You can get those from Boost Math. Look at the documentation of Boost Math:
https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/dist.html for general advice and
https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/dist_ref/dists/beta_dist.html for the beta distribution.
The docs have many examples.

HDembinski: With Boost Math, the implementation should be easy, you just need to call the quantile with the arguments listed here https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Clopper%E2%80%93Pearson_interval

include/boost/histogram/utility/agresti_coull_interval.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  explicit agresti_coull_interval(deviation d = deviation{1.0}) noexcept
+      : z_{static_cast<double>(d)} {}
+
+  interval_type operator()(value_type successes, value_type failures) const noexcept {
+    const value_type half{0.5}, one{1.0}, zsq = z_ * z_;
+    const value_type ns = successes, nf = failures;
+    const value_type n = ns + nf;
+    const value_type n_ac = n + zsq;
+    const value_type a = (ns + (zsq * half)) / n_ac;
```
HDembinski, line 34: Please remove the superfluous parentheses.

include/boost/histogram/utility/jeffreys_interval.hpp
```diff
@@ -0,0 +1,55 @@
...
+template <class ValueType>
+class jeffreys_interval : public binomial_proportion_interval<ValueType> {
+  using base_t = binomial_proportion_interval<ValueType>;
+
+public:
+  using value_type = typename base_t::value_type;
+  using interval_type = typename base_t::interval_type;
+
+  explicit jeffreys_interval(deviation d = deviation{1.0}) noexcept
+      : z_{static_cast<double>(d)} {}
```
HDembinski, line 27: Like for the Clopper-Pearson interval, you rather want to accept `confidence_level` here instead of `deviation`.

include/boost/histogram/utility/jeffreys_interval.hpp
```diff
@@ -0,0 +1,55 @@
...
+    const value_type alpha1 = (one - d) * half;
+    const value_type alpha2 = one - ((one - d) * half);
+    // Source: https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Clopper%E2%80%93Pearson_interval
+    const value_type m = ns + half;
+    const value_type n = n - ns + half;
+    // Source: https://en.wikipedia.org/wiki/Beta_distribution
+    const value_type beta = (tgamma(m)*tgamma(n)) / tgamma(m+n);
+    // Source: https://repository.upenn.edu/cgi/viewcontent.cgi?article=1440&context=statistics_papers
+    const value_type a = (one / beta) * std::pow(alpha1, m - one) * std::pow(1 - alpha1, n - one);
+    const value_type b = (one / beta) * std::pow(alpha2, m - one) * std::pow(1 - alpha2, n - one);
```
HDembinski, line 43: Like the implementation of Clopper-Pearson, this is probably wrong. Also here, you need the quantiles of the beta distribution. See my comments in `clopper_pearson_interval`.

include/boost/histogram/utility/wald_interval.hpp
```diff
@@ -0,0 +1,45 @@
...
+  using value_type = typename base_t::value_type;
+  using interval_type = typename base_t::interval_type;
+
+  explicit wald_interval(deviation d = deviation{1.0}) noexcept
+      : z_{static_cast<double>(d)} {}
+
+  interval_type operator()(value_type successes, value_type failures) const noexcept {
+    const value_type ns = successes, nf = failures;
+    const value_type n = ns + nf;
+    const value_type a = (ns) / (n);
```
HDembinski, line 32: Please remove the superfluous parentheses.

include/boost/histogram/utility/wald_interval.hpp
```diff
@@ -0,0 +1,45 @@
...
+    const value_type b = (z_ / (n) * std::sqrt(n)) * std::sqrt(ns * nf);
```
HDembinski, line 33: It is numerically more stable (after faster in this case, too) you group all the computations under the square roots together. You can do `sqrt(ns * nf / n)`.

include/boost/histogram/utility/wilson_interval.hpp
```diff
@@ -0,0 +1,54 @@
...
+  using interval_type = typename base_t::interval_type;
+
+  explicit wilson_interval(deviation d = deviation{1.0}) noexcept
+      : z_{static_cast<double>(d)} {}
+
+  interval_type operator()(value_type successes, value_type failures) const noexcept {
+    const value_type half{0.5}, quarter{0.25}, zsq = z_ * z_;
+    const value_type ns = successes, nf = failures;
+    const value_type n = ns + nf;
+    const value_type a = (ns + half * zsq) / (n + zsq);
```
HDembinski, line 41: Yes, here I made the mistake, too, it should be `double` instead of `value_type` here.

include/boost/histogram/utility/wilson_interval.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+public:
+  using value_type = typename base_t::value_type;
+  using interval_type = typename base_t::interval_type;
+
+  explicit wilson_interval(deviation d = deviation{1.0}) noexcept
+      : z_{static_cast<double>(d)} {}
+
+  interval_type operator()(value_type successes, value_type failures) const noexcept {
+    const value_type half{0.5}, quarter{0.25}, zsq = z_ * z_;
```
HDembinski, line 38: This is wrong, the type should be `double` instead of `value_type` here.

include/boost/histogram/utility/binomial_proportion_interval.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <cmath>
+#include <stdexcept>
+
+namespace boost {
+namespace histogram {
+namespace utility {
+
+template <class ValueType>
+class binomial_proportion_interval {
+public:
```
HDembinski, line 21: @gohil-jay Please add a `static_assert` here to constrain the `ValueType` to floating point numbers. You can use the template functions in the header `type_traits` from the standard library to check whether `ValueType` is a floating point type.

test/accumulators_fraction_test.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+template <class T>
+void run_tests() {
+  using f_t = fraction<T>;
+
+  // ctor
+  {
+    f_t f(1, 2);
+    BOOST_TEST_EQ(f.successes(), 1);
+    BOOST_TEST_EQ(f.failures(), 2);
```
HDembinski, line 24: Jay, I observe a tendency of you to tackle the difficult tasks before the easy tasks. Remember, I asked you to add tests here before implementing the intervals. Can you please add some tests here?

include/boost/histogram/utility/clopper_pearson_interval.hpp
gohil-jay: I should rename it, it's 1st arguments for low and high variables (α/2 and 1-α/2). And yes I interchanged d with cl, so I'll α from cl/2 as you mentioned!

include/boost/histogram/utility/clopper_pearson_interval.hpp
gohil-jay: It's for calculating B(α, β) for f(x;α,β) for the intervals; = (α+1)!*(β+1)! / (α+β+1)! and tgamma(x) = (x-1)!

(Can explain in a meet)

HDembinski: This is not calculating B(alpha, beta). B(alpha, beta) is the quantile of a pdf which does not have a closed form. You cannot implement the formula this way. In fact, you cannot implement the formula in code at all. Please read the references that I linked to.

HDembinski: Your only option is to use the beta distribution from Boost Math, so please do that instead of trying other solutions that do not work.

test/accumulators_fraction_test.cpp
gohil-jay: Will do this next! Already started.

include/boost/histogram/utility/agresti_coull_interval.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+public:
+  using value_type = typename base_t::value_type;
+  using interval_type = typename base_t::interval_type;
+
+  explicit agresti_coull_interval(deviation d = deviation{1.0}) noexcept
+      : z_{static_cast<double>(d)} {}
+
+  interval_type operator()(value_type successes, value_type failures) const noexcept {
+    const double half{0.5}, one{1.0}, zsq = z_ * z_;
```
HDembinski, line 30: if you use declare them to be `double`, you don't need the constants `half` and `one`, you can just use the numbers in the formula directly. But after switching back and forth, I think it is correct to replace `double` with `value_type` here.

What we really want is to ensure that `value_type` is a floating_point number. The `value_type` of the interval is not necessarily the same value_type that the class `fraction` uses. The latter can be an integer, but the value_type of the interval classes must be of floating point type.

HDembinski: This is already ensured by the base class `binomial_proportion_interval`.

include/boost/histogram/accumulators/fraction.hpp
```diff
@@ -57,8 +57,8 @@ class fraction {
   }
 
 private:
-  value_type succ_ = 0;
-  value_type fail_ = 0;
+  double succ_ = 0;
+  double fail_ = 0;
```
HDembinski, line 61: This replacement is wrong, the others are correct. Please revert.

include/boost/histogram/utility/agresti_coull_interval.hpp
```diff
@@ -27,12 +27,12 @@ class agresti_coull_interval : public binomial_proportion_interval<ValueType> {
...
   interval_type operator()(value_type successes, value_type failures) const noexcept {
-    const value_type half{0.5}, one{1.0}, zsq = z_ * z_;
+    const double half{0.5}, one{1.0}, zsq = z_ * z_;
     const value_type ns = successes, nf = failures;
     const value_type n = ns + nf;
     const value_type n_ac = n + zsq;
-    const value_type a = (ns + (zsq * half)) / n_ac;
-    const value_type b = z_ * std::sqrt((one - a) * a / n_ac);
+    const double a = (ns + zsq * half) / n_ac;
+    const double b = z_ * std::sqrt((one - a) * a / n_ac);
```
HDembinski, line 35: This replacement is wrong, please revert.

include/boost/histogram/utility/agresti_coull_interval.hpp
HDembinski: In other words, please revert this change.

include/boost/histogram/utility/clopper_pearson_interval.hpp
```diff
@@ -23,32 +23,36 @@ class clopper_pearson_interval : public binomial_proportion_interval<ValueType>
...
   using interval_type = typename base_t::interval_type;
 
-  explicit clopper_pearson_interval(deviation d = deviation{1.0}) noexcept
-      : z_{static_cast<double>(d)} {}
+  explicit clopper_pearson_interval(confidence_level cl = confidence_level{0.683}) noexcept
+      : cl_{static_cast<double>(cl)} {}
 
   interval_type operator()(value_type successes, value_type failures) const noexcept {
-    const value_type half{0.5}, one{1.0};
+    const double half{0.5}, one{1.0};
```
HDembinski, line 31: Please revert this.

include/boost/histogram/utility/clopper_pearson_interval.hpp
```diff
@@ -23,32 +23,36 @@ class clopper_pearson_interval : public binomial_proportion_interval<ValueType>
...
     const value_type n2 = n - ns;
     // Source: https://en.wikipedia.org/wiki/Beta_distribution
-    const value_type beta1 = (tgamma(m1)*tgamma(n1)) / tgamma(m1+n1);
-    const value_type beta2 = (tgamma(m2)*tgamma(n2)) / tgamma(m2+n2);
+    const double beta1 = (tgamma(m1)*tgamma(n1)) / tgamma(m1+n1); // Calculating B(α, β) for f(x;α,β) for low_interval, = α!*β! / (α+β)!
+    const double beta2 = (tgamma(m2)*tgamma(n2)) / tgamma(m2+n2); // Calculating B(α, β) for f(x;α,β) for high_interval, = α!*β! / (α+β)!
+    // Calculating beta distribution values for the low_interval and high_interval | f(x;α,β) = 1/B(α, β) * x^(α-1) * (1-x)^(β-1)
     // Source: https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Clopper%E2%80%93Pearson_interval
-    const value_type a = (one / beta1) * std::pow(alpha1, m1 - one) * std::pow(1 - alpha1, n1 - one);
-    const value_type b = (one / beta2) * std::pow(alpha2, m2 - one) * std::pow(1 - alpha2, n2 - one);
```
HDembinski: Use `value_type` instead of `double` everywhere.

include/boost/histogram/utility/clopper_pearson_interval.hpp
```diff
@@ -23,32 +23,36 @@ class clopper_pearson_interval : public binomial_proportion_interval<ValueType>
...
+    const double a = (one / beta1) * std::pow(alpha1, m1 - one) * std::pow(1 - alpha1, n1 - one);
+    const double b = (one / beta2) * std::pow(alpha2, m2 - one) * std::pow(1 - alpha2, n2 - one);
     return std::make_pair(a, b);
   }
 
 private:
-  double z_;
+  double cl_;
```
HDembinski, line 49: should also be `value_type`

include/boost/histogram/utility/wilson_interval.hpp
```diff
@@ -35,11 +35,11 @@ class wilson_interval : public binomial_proportion_interval<ValueType> {
...
 
   interval_type operator()(value_type successes, value_type failures) const noexcept {
-    const value_type half{0.5}, quarter{0.25}, zsq = z_ * z_;
+    const double half{0.5}, quarter{0.25}, zsq = z_ * z_;
     const value_type ns = successes, nf = failures;
     const value_type n = ns + nf;
-    const value_type a = (ns + half * zsq) / (n + zsq);
-    const value_type b = z_ / (n + zsq) * std::sqrt(ns * nf / n + quarter * zsq);
+    const double a = (ns + half * zsq) / (n + zsq);
+    const double b = z_ / (n + zsq) * std::sqrt(ns * nf / n + quarter * zsq);
```
HDembinski, line 42: please revert

include/boost/histogram/utility/wald_interval.hpp
```diff
@@ -29,8 +29,8 @@ class wald_interval : public binomial_proportion_interval<ValueType> {
   interval_type operator()(value_type successes, value_type failures) const noexcept {
     const value_type ns = successes, nf = failures;
     const value_type n = ns + nf;
-    const value_type a = (ns) / (n);
-    const value_type b = (z_ / (n) * std::sqrt(n)) * std::sqrt(ns * nf);
+    const double a = ns / n;
+    const double b = (z_ / n) * std::sqrt(ns * nf / n);
```
HDembinski, line 33: use value_type not double, otherwise good

test/Jamfile
```diff
@@ -48,7 +48,7 @@ alias cxx14 :
       <toolset>clang:<cxxflags>"-O3 -ffast-math" ]
     [ run accumulators_weighted_mean_test.cpp ]
     [ run accumulators_weighted_sum_test.cpp ]
-    [ run algorithm_project_test.cpp ]
+    # [ run algorithm_project_test.cpp ]
```
HDembinski, line 51: this should not be necessary

test/odr_test.cpp
```diff
@@ -17,7 +17,11 @@
...
 #include <boost/histogram/detail/ignore_deprecation_warning_begin.hpp>
 #include <boost/histogram/detail/ignore_deprecation_warning_end.hpp>
+#include <boost/histogram/utility/binomial_proportion_interval.hpp>
 #include <boost/histogram/utility/agresti_coull_interval.hpp>
 #include <boost/histogram/utility/clopper_pearson_interval.hpp>
 #include <boost/histogram/utility/jeffreys_interval.hpp>
 #include <boost/histogram/utility/wald_interval.hpp>
+#include <boost/histogram/utility/wilson_interval.hpp>
+#include <boost/histogram/detail/normal.hpp>
+#include <boost/histogram/detail/erf_inv.hpp>
```
HDembinski, line 27: these changes are not necessary, please revert

HDembinski:
The new intervals start to look good but you started to make changes that broke things that were previously correct. I am talking especially about the changes to `fraction` and `odr_test.cpp`

test/Jamfile
HDembinski: If you are still seeing this test failing, you need to use `cxxstd=17` instead of `cxxstd=latest` in your b2 call

HDembinski: This is a bug in the microsoft compiler

include/boost/histogram/utility/wald_interval.hpp
gohil-jay: Umm... But what if it's an int ?

test/Jamfile
gohil-jay: Noted.

include/boost/histogram/utility/wald_interval.hpp
gohil-jay: Got it nevermind.

test/accumulators_fraction_test.cpp
```diff
@@ -19,15 +19,25 @@ void run_tests() {
...
   {
-    f_t f(1, 2);
-    BOOST_TEST_EQ(f.successes(), 1);
-    BOOST_TEST_EQ(f.failures(), 2);
+    f_t f1(3, 1);
+    f_t f2(3.0, 5.0);
+    f_t f3(1.0, 3.0);
+    BOOST_TEST_EQ(f1.successes(), 3); BOOST_TEST_EQ(f1.failures(), 1); BOOST_TEST_EQ(f1.successes(), 3.0); BOOST_TEST_EQ(f1.failures(), 1.0);
+    BOOST_TEST_EQ(f2.successes(), 3); BOOST_TEST_EQ(f2.failures(), 5); BOOST_TEST_EQ(f2.successes(), 3.0); BOOST_TEST_EQ(f2.failures(), 5.0);
+    BOOST_TEST_EQ(f3.successes(), 1); BOOST_TEST_EQ(f3.failures(), 3); BOOST_TEST_EQ(f3.successes(), 1.0); BOOST_TEST_EQ(f3.failures(), 3.0);
```
HDembinski, line 27: You do not need to compare to 3 **and** 3.0,  etc., this is guaranteed by C++. Please remove the superfluous tests.

HDembinski: Otherwise this looks good.

test/accumulators_fraction_test.cpp
```diff
@@ -19,15 +19,25 @@ void run_tests() {
...
+    BOOST_TEST_EQ(f1.value(), 0.75);
+    BOOST_TEST_EQ(f2.value(), 0.375);
+    BOOST_TEST_EQ(f3.value(), 0.25);
+    BOOST_TEST_EQ(f1.variance(), 12); BOOST_TEST_EQ(f1.variance(), 12.0);
+    BOOST_TEST_EQ(f2.variance(), 120); BOOST_TEST_EQ(f2.variance(), 120.0);
+    BOOST_TEST_EQ(f3.variance(), 12); BOOST_TEST_EQ(f3.variance(), 12.0);
+    // BOOST_TEST_EQ(f1.confidence_interval().first, 0.5); BOOST_TEST_EQ(f1.confidence_interval().second, 0.9);
```
HDembinski, line 34: You need BOOST_TEST_IS_CLOSE here.

test/accumulators_fraction_test.cpp
```diff
@@ -19,15 +19,25 @@ void run_tests() {
...
   }
 }
 
 int main() {
 
-  run_tests<int>();
+  // run_tests<int>(); // confidence_interval() throws error due to static_assert in binomial_proportion_interval()
```
HDembinski, line 40: This test exposes an error in the implementation in `fraction`. If `value_type` is an integer, which is perfectly valid, then the confidence interval computer should still return an interval in `double`. To get this working correctly requires a bit of metaprogramming that we can resolve together tomorrow in our call.

test/accumulators_fraction_test.cpp
```diff
@@ -28,6 +28,23 @@ void run_tests() {
...
+  }
+
+  {
+    using f_t1 = fraction<double>;
+    using f_t2 = fraction<int>;
+    f_t1 f1(5,3);
+    f_t2 f(f1.successes(), f1.failures());
+    BOOST_TEST_EQ(f.successes(), 5);
+    BOOST_TEST_EQ(f.failures(), 3);
+  }
```
HDembinski, line 46: that's not how the converting constructor works. What you are testing here always works, even without making a special constructor. You need to test `f_t1 f1(5, 3); f_t2 f2(f1);`

HDembinski: The other test you added looks good. :+1:

HDembinski:
@gohil-jay Can you work on fixing the failing tests? appveyor is complaining about possible loss of data
```
C:\projects\boost\boost/histogram/accumulators/fraction.hpp(35): error C2220: the following warning is treated as an error
test\accumulators_fraction_test.cpp(43): note: see reference to function template instantiation 'boost::histogram::accumulators::fraction<T>::fraction<double>(const boost::histogram::accumulators::fraction<double> &) noexcept' being compiled
        with
        [
            T=int
        ]
test\accumulators_fraction_test.cpp(43): note: see reference to function template instantiation 'boost::histogram::accumulators::fraction<T>::fraction<double>(const boost::histogram::accumulators::fraction<double> &) noexcept' being compiled
        with
        [
            T=int
        ]
test\accumulators_fraction_test.cpp(77): note: see reference to function template instantiation 'void run_tests<int>(void)' being compiled
C:\projects\boost\boost/histogram/accumulators/fraction.hpp(35): warning C4244: 'argument': conversion from 'double' to 'const int', possible loss of data
C:\projects\boost\boost/histogram/accumulators/fraction.hpp(35): warning C4244: 'argument': conversion from 'double' to 'const int', possible loss of data
```

test/utility_clopper_pearson_interval_test.cpp
```diff
@@ -19,17 +19,17 @@ int main() {
...
+    const auto x = iv(1, 1);
+    BOOST_TEST_IS_CLOSE(x.first, 0.0, deps);
+    BOOST_TEST_IS_CLOSE(x.second, 0.8413447460685429, deps);
+  }
+
+  {
+    const auto x = iv(1, 1);
+    BOOST_TEST_IS_CLOSE(x.first, 0.1586552539314571, deps);
+    BOOST_TEST_IS_CLOSE(x.second, 1.0, deps);
+  }
```
HDembinski, line 32: Why this change? You should not change the original test condition, because that should have been correct.

doc/Jamfile
```diff
@@ -40,7 +40,7 @@ doxygen reference
 
 actions doxygen-postprocessing
 {
-  python3 $(THIS_PATH)/doxygen_postprocessing.py "$(>)"
+  python $(THIS_PATH)/doxygen_postprocessing.py "$(>)"
```
HDembinski, line 43: We cannot change this yet, because on some platforms `python` still points to python2.

test/accumulators_fraction_test.cpp
```diff
@@ -0,0 +1,82 @@
...
+    BOOST_TEST_EQ(f.successes(), 3);
+    BOOST_TEST_EQ(f.failures(), 5);
+    BOOST_TEST_EQ(f.value(), 0.375);
+    BOOST_TEST_IS_CLOSE(f.variance(), 0.375 * (1 - 0.375) / 8, eps);
+
+    const auto ci = f.confidence_interval();
+    const auto expected = utility::wilson_interval<double>()(3, 5);
+    BOOST_TEST_IS_CLOSE(ci.first, expected.first, eps);
+    BOOST_TEST_IS_CLOSE(ci.second, expected.second, eps);
+  }
```
HDembinski, line 72: This test seems superfluous. You should rather check what happens in the special case when both successes and failures is zero.

test/utility_clopper_pearson_interval_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <limits>
+#include "is_close.hpp"
+#include "throw_exception.hpp"
+
+using namespace boost::histogram::utility;
+
+int main() {
+
+  // const double deps = std::numeric_limits<double>::epsilon();
+  const double deps = 0.1;
```
HDembinski, line 18: This requires a comment why deps is so large.

test/utility_jeffreys_interval_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <limits>
+#include "is_close.hpp"
+#include "throw_exception.hpp"
+
+using namespace boost::histogram::utility;
+
+int main() {
+
+  // const double deps = std::numeric_limits<double>::epsilon();
+  const double deps = 0.1;
```
HDembinski, line 18: This requires a comment why deps has to be so large.

HDembinski:
Hi Jay, the comments that you added look ok. I am concerned about the test that you changed, and that we do not have an explanation for the large deviations of the jeffreys and clopper pearson intervals from the reference numbers. It is ok to have deviations of about 1e-7, but not 0.1.

HDembinski:
The error in Slow/gcc10 is not caused by us, but in boostorg/math. I send a PR with a fix to boostorg/math.

HDembinski:
@gohil-jay I fixed the jefferys_interval, you don't need to change this anymore. While reading up on it, I found a minor enhancement which I implemented. It is all documented in the class.

While fixing the jeffreys_interval, I noticed why you were getting wrong results. You treated CL as alpha, but alpha = 1 - CL. This is also the reason why your implementation of the clopper_pearson interval gives wrong results. We can and should precompute alpha_half in the constructor of the interval class, because alpha_half is constant during the life-time of the interval maker. This is why we use a class, to be able to put computations that only need to be done once into the constructor. Please have a look how I did it in the jeffreys_interval class, and copy that for the clopper_pearson_interval.

HDembinski:
I now also fixed the clopper-pearson interval. You don't need to change anything there anymore. Since the beta distribution from boost::math cannot handle the case where the successes or failures are zero, I use an explicit formula for that special case, which is described in the wikipedia.

test/utility_clopper_pearson_interval_test.cpp
gohil-jay: This is a mistake on my end, I changed this while finding where the issue came from (and realized it wasn't this) and missed changing it back. I see you've fixed it.

test/accumulators_fraction_test.cpp
gohil-jay: Agreed. Interestingly it should show errors for 0/0 among others. I will look into it.

test/utility_clopper_pearson_interval_test.cpp
gohil-jay: I see your improved implementation now covers 0.001. I will take a look on how you did it.

test/utility_jeffreys_interval_test.cpp
gohil-jay: I see your improved implementation now covers 0.001. I will take a look on how you did it.

gohil-jay:
I will take a look at all the changes including one to boost::math to see how you overcame beta_distribution limit (maybe conditional implementation for those inputs or something I missed from docs) and rest of others.

test/utility_clopper_pearson_interval_test.cpp
HDembinski: You used alpha = CL =0.68. instead of alpha = 1 - CL =0.32.

HDembinski:
> I will take a look at all the changes including one to boost::math to see how you overcame beta_distribution limit (maybe conditional implementation for those inputs or something I missed from docs) and rest of others.

Yes, I special-cased this. For successes = 0 (or failures = 0), one can compute the interval boundaries analytically as described in the Wikipedia.

HDembinski:
@henryiii Can you please look over the interface of accumulators/fraction.hpp and let us know if you see any issues? The interface to compute `confidence_interval` is intentionally not flexible, since people should use the wilson interval unless they really know what they are doing. We provide more ways to compute intervals with the interval computers in `boost/histogram/utility`.

henryiii:
Is there an integration test somewhere (maybe I’m just missing it)? I see unit tests for pieces, but is there a test were a histogram with the new accumulator is tested? I wanted to see the bool sample being used, I’m thinking this is the first storage where a bool sample is expected. That's the most important usage from boost-histogram's perspective.

No docs yet?

To be sure the interface works in boost-histogram, we can always make a branch of boost-histogram building on this branch. But I don’t see anything that would be likely to be a problem.

HDembinski:
@henryiii Thank you for the comments.

> Is there an integration test somewhere (maybe I’m just missing it)? I see unit tests for pieces, but is there a test were a histogram with the new accumulator is tested? I wanted to see the bool sample being used, I’m thinking this is the first storage where a bool sample is expected. That's the most important usage from boost-histogram's perspective.

Good catch, I think we tested this in the beginning interactively on godbolt, but we don't have this integration test. Needs to be added.
 
> No docs yet?

Do you mean the user guide? All new classes/functions are documented with doxygen strings. I also checked the generated docs, they look ok. But it is still missing from the guide, that's true.

> To be sure the interface works in boost-histogram, we can always make a branch of boost-histogram building on this branch. But I don’t see anything that would be likely to be a problem.

Yes and until the next boost release, we can also make breaking changes.

henryiii:
> Do you mean the user guide?

Yes. Not at all a blocker, can be added later, just should be on a todo list somewhere.

henryiii:
Looks great. Boost-histogram doesn't actually use std::vector (except for strings), so I think we'll be fine there.

HDembinski:
Docs were improved, integration test added, and some convenience to allow the interval computers to accept a fraction accumulator directly. I will merge as soon as this passes all tests. Thanks @gohil-jay @henryiii

PR #365, opened 2022-09-17
fix cmake build of benchmarks

HDembinski:


HDembinski:
I am closing this, because the fix was already implemented in another branch. The failure has been reported to MS, because it is likely caused by a change in the compiler.

PR #384, opened 2023-02-18
Replace detail::span and detail::make_span with implementations in boost::core

HDembinski:


HDembinski:
@glenfe It works like a charm now. Thanks!

glenfe:
@HDembinski great.   Is `boost::make_span` (from `<boost/core/make_span.hpp>`) OK too?

HDembinski:
> @HDembinski great. Is `boost::make_span` (from `<boost/core/make_span.hpp>`) OK too?

Thanks for the pointer; I missed that you also added make_span, works also fine!

HDembinski:
> @HDembinski great. Is `boost::make_span` (from `<boost/core/make_span.hpp>`) OK too?

Thanks for the pointer; I missed that you also added make_span, works also fine!

glenfe:
Looks good; thanks @HDembinski

Issue #367, opened 2022-09-28
Extend fraction to weighted samples


HDembinski:
We have a fraction accumulator now, but it does not work correctly for weighted samples. However, we definitely need this in high-energy physics because we often estimate the efficiency of some detector from the weighted simulation samples.

There are several ways how this could be implemented
1) New class `weighted_fraction<T>`
2) Specialization `fraction<weighted_sum<T>>`
3) Template specialization `fraction<T, bool weight_support = False>`

3) is my least favored. I like 2), if that's feasible.

HDembinski:
For those interest, the math for weighted fractions is discussed here https://arxiv.org/abs/2110.00294

PR #385, opened 2023-06-05
Add support for weighted fraction accumulator

ryanelandt:
This PR extends `fraction`to weighted samples to address issue #367.

I took the approach of creating a new `weighted_fraction` class. This class is a composition of two classes: `fraction` and a new internal class `sum_of_weights_squared`. I explain the thought process behind these choices.

The classes `sum`, `mean`, and `fraction` are non-weighted. Each of these classes could almost function as a weighted class, but requires one additional piece of information: the sum of the weights squared, that is to say $\sum (w^2)$. So, I turned this piece of information into a class `sum_of_weights_squared`. I created the class `weighted_fraction` by putting `fraction` and `sum_of_weights_squared` together.

The code passes all unit tests, but has `Wpedantic` warnings because `wilson_solve` uses a designated initializer (i.e., `wilson_solve({.n_eff=n_eff, .p_hat=p_hat, .correction=correction})`). There's likely a good way to avoid getting this warning, while also enforcing input meaning, but I'm missing it.


HDembinski:
Thanks for this substantial contribution, I will review it next weekend. I think you can replace the initializer to remove the warning.

include/boost/histogram/accumulators/fraction.hpp
```diff
@@ -36,7 +37,8 @@ class fraction {
   using const_reference = const value_type&;
   using real_type = typename std::conditional<std::is_floating_point<value_type>::value,
                                               value_type, double>::type;
-  using interval_type = typename utility::wilson_interval<real_type>::interval_type;
+  using score_type = typename utility::wilson_interval<real_type>;
```
HDembinski, line 40: score_type should not appear here, because it is not a type that needs to be publicly visible.

ryanelandt: This issue should be addressed now. Please take a look.

ryanelandt:
The failing check is due to the job never starting and eventually timing out.

HDembinski:
This sometimes happens. I restarted the job. If it continues to fail, we can ignore it. Checking compatibility with an ancient compiler becomes less and less useful as time progresses.

ryanelandt:
Please let me know if there are any changes I can make to this.

ryanelandt:
I moved the `weighted_fraction` class to an `experimental` folder. I also made `weighted_fraction` a friend of `fraction` which helps minimize changes to the public API. Let me know if there are any other changes I can make.

PR #386, opened 2023-06-13
Piecewise

### LINKED ISSUE #47: New axis: piecewise_regular
A composed axis of several connected regular axis without gaps. Should use regular_axis underneath.

ryanelandt:
This PR hopes to eventually close #47, adding a piecewise axis. The structure of this piecewise axis implementation is different from the others. I'm hoping to get feedback on it. Changes to this PR, and supporting PRs will likely be needed before this PR can be merged in. 

### Solution overview:

This 1D piecewise axis implementation has two layers. The first layer has to do with creating immutable pieces that transform from input space X to bin space Y. The second layer modifies the behavior of these transformations in order to make them behave like an axis. If axis growth is allowed, this second layer is mutable. I explain how this works on a high level.

## Layer 1: Pieces

Suppose we want to divide the input space X into 4 bins of unit spacing starting at x=3.
```
bin   0   1   2   3
    |---|---|---|---|
   x=3             x=7
```
This piece is described by three pieces of information
  - forward transformation: y = f(x) = x - 3
  - inverse transformation: x = f⁻¹(y) = y + 3
  - number of bins: N = 4

There are five supported piece types:
  1. unit      (bin spacing is 1)
  2. uniform   (bin spacing is constant)
  3. multiply  (bin spacing is multiplied by a constant)
  4. add       (bin spacing is added to by a constant)
  5. variable (bin bounds are user supplied points)

**Note:** Pieces are similar to the existing transformation functionality in the sqrt and log transformations of regular axis. The difference is that pieces calculate bins, and be centered at arbitrary locations. 

### Combining pieces

Suppose we wanted to combine the following two pieces.
```
  bin   0   1   2   3          bin  0  1  2   3    4
      |---|---|---|---|            |-|--|---|----|-----|
     x=3             x=7          x=9                x=15
         Unit spacing                   Add spacing
```
Typically one wouldn't want to combine two pieces with a gap between them, but this possibility is supported by the implementation. A piecewise transformation stores each of these unmodified pieces internally. The external behavior is that of piece with bins 0 to 8 shown below.
```
  bin   0   1   2   3     nan    4  5  6   7    8
      |---|---|---|---|         |-|--|---|----|-----|
     x=3             x=7       x=9                x=15
```

### Bin shifting:

Suppose we wanted to use the piecewise transformation above in a growable axis. This axis starts off with 9 bins. Two kinds of bins can be added: underflow and overflow. Underflow bins are added below x=3. Overflow bins are added above x=15. Adding bins between x=7 and x=9 is not supported. The mutable piece `x_bin_shift` accumulates:
  - added underflow bins, and
  - added overflow bins.
 
Using this information, x_bin_shift augments the behavior of the contained piecewise transformation as follows. The size (i.e., number of bins) of the piece is increased by the number of underflow and overflow bins. The bin values (i.e., Y axis) is shifted by the number of underflow bins in the forward and inverse calculations.

### Linear axis:

Wrapping the, possibly shifted, piece in an `int_resolver_linear` makes the piece behave like an axis. This linear int resolver implements an index method.

### Circular axis:

Suppose we wanted to make the above described piecewise transformation circular with a relevant period x=3 to x=23. This piece is copied below for convenience.
```
  bin   0   1   2   3     nan    4  5  6   7    8
      |---|---|---|---|         |-|--|---|----|-----|
     x=3             x=7       x=9                x=15
```
To do this, we wrap the piece in an `int_resolver_circular` with the bounds x=3 and x=23. This circular resolver augments the behavior of the contained piece. The input (i.e., the X axis) is wrapped to the domain x=3 to x=23 before being given to the contained piece This circular int resolver implements an index method allowing it to act as an axis.


ryanelandt:
Integrating int_resolver functionality into regular axis seems like a good idea for a first step. To do this, I added PR #387. This PR eliminates some of the precision issues with regular axis. It includes a before and after indexing benchmark.

My initial post for this PR (#386) did not clearly communicate that gaps between axes are handled with a single overflow bin. I’ve edited it to correct this.

PR #387, opened 2023-06-15
Improve axis regular precision

ryanelandt:
This PR refactors regular axis to address the issues described in #336 by incorporating some of the ideas from #386. This PR adds a unit that verifies exact indexing for various cases.

The output of `./benchmark_axis_index` before the refactor:  
```
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
regular<uniform>               1.31 ns         1.31 ns    538240574
regular<normal>                2.00 ns         2.00 ns    351784388
circular<uniform>              1.62 ns         1.62 ns    432490637
circular<normal>               1.62 ns         1.62 ns    424151491
```

The output of `./benchmark_axis_index` after the refactor:
```
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
regular<uniform>               1.13 ns         1.13 ns    578725526
regular<normal>                1.71 ns         1.71 ns    399524265
circular<uniform>              1.51 ns         1.51 ns    471027721
circular<normal>               1.49 ns         1.49 ns    459937000
```

### DIFF
--- a/examples/getting_started_listing_01.cpp
+++ b/examples/getting_started_listing_01.cpp
@@ -72,14 +72,16 @@ int main() {
     Edges can be accessed with methods `lower()` and `upper()`.
   */
 
-  // IEEE 754 has a positive and a negative zero, which are equal, but have different signs.
-  // This lambda function converts negative zero to positive zero for the purposes of string formatting.
+  // IEEE 754 has a positive and a negative zero, which are equal, but have
+  // different signs. This lambda function converts negative zero to positive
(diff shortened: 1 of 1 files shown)


emmenlau:
This looks actually quite promising, and performance is better than with our (naiive) solution. Could this be considered for integration?

henryiii:
By the way, I think this would help https://github.com/scikit-hep/hist/issues/455.

PR #388, opened 2023-06-29
Split out dynamic-only tests

ecatmur:
Compiling tests under Visual Studio 2019 14.29.30133 (19.29.30145.0) fails with a not very helpful error:
```
libs\histogram\test\algorithm_project_test.cpp(186,1): fatal error C1903: unable to recover from previous error(s); stopping compilation
```
This can be fixed by moving the dynamic-only tests into a separate function, so that they are only compiled and run once.

### DIFF
--- a/test/algorithm_project_test.cpp
+++ b/test/algorithm_project_test.cpp
@@ -136,6 +136,8 @@ void run_tests() {
   }
 }
 
+// split out dynamic tests as workaround for compiler bug in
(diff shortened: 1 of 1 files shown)


HDembinski:
@ecatmur Thank you for this nice patch and sorry for the long delay in responding. I added a comment to the code why this was changed. How did you find out this fix?

Issue #349, opened 2022-01-19
Add collector accumulator


HDembinski:
A simple accumulator that just fills a `std::vector` with the samples passed to it. The user can then run any kind of statistical estimation on the sample. It is a very inefficient accumulator, but sometimes that's acceptable and the flexibility it offers is more important than the performance.

Pseudo-code
```c++
// C++17
namespace bh = boost::histogram;
auto h = bh::make_histogram(bh::axis::integer(0, 2),
              bh::dense_storage<bh::accumulators::Collector<>>());

// fill
h(0, 1);
h(0, 2);
h(0, 3);
h(1, 4);
h(1, 5);

for (auto x: h.at(0))
   std::cout << x << std::endl;
```
Prints: 1 2 3

wiso:
I think this would be a very nice feature since this is basically a groupby. I have submitted a simple PR.

wiso:
I guess this can be closed due to #390

HDembinski:
True.

PR #390, opened 2023-11-06
Add collector accumulator

wiso:
This is a simple implementation of collector accumulator, a very simple accumulator which stores all the values passed, as suggested in #349. Some new tests are implemented. Documentation is updated.

I would like to have a code review. For sure these items are missing.

- [x] add copyright
- [x] check compilation with b2

Probably implementing serialization for this is not a good idea.

I had to change the continuous integration:
   * fix how libraries are searched by `depinst.py`, since by default the `example` directory is not used
   * move to Ubuntu 20.04 since 18.04 is no longer supported https://github.com/actions/runner-images/issues/6002. Because of that also moved from gcc5 to gcc7.


wiso:
I am  not sure why the compilation problems. This works:

```
docker run -it --rm -v $HOME/histogram:/histogram ubuntu:latest
cd /histogram
mkdir build
cd build
apt-get update
apt-get install git cmake build-essential

cmake -DCMAKE_BUILD_TYPE=Debug ..
cmake --build . --target check -j3
```

wiso:
is this project abandoned? @HDembinski

HDembinski:
Not at all, but I have a lot of things on my plate right now. I should be able to work on all the outstanding issues/PRs in the coming months.

HDembinski (changes requested):
Thanks for the effort, this looks very good. The implementation looks correct and the tests thorough.

Before merging, I would like to explore whether this code can be generalized in such a way that one can pass several sample values to the collector. Thanks to the template system, this should work with this implementation already if the size of the value array is known at compile file. In other words, if one uses a `std::array` as ValueType. A test for this case should be added.

However, I also see the need for a dynamic collector, where the size of the sample array is only known at runtime. While this can be handled with a `ValueType = std::vector<double>`, it is very inefficient since it will cause a lot of dynamic memory allocations.

Perhaps it is necessary to write a separate class, called dynamic_collector, to implement this. Then this PR can be merged right now. But if the two implementations can be merged and use the same template class name, that would be nice.

Implementing a dynamic collector efficiently could be achieved with a new container type which stores tuples of numbers. The size of the tuple is defined at construction time. Internally, it would be a `std::vector`, the tuples are stored consecutively in memory.

HDembinski:
Please merge current develop branch to fix the CI issues.

HDembinski:
You don't need to write a changelog message, that's why I removed that item from your todo list. Please add the Boost copyright notice to all new files.

wiso:
what should I write? I see a common

```
//
// Distributed under the Boost Software License, version 1.0.
// (See accompanying file LICENSE_1_0.txt
// or copy at http://www.boost.org/LICENSE_1_0.txt)
```

What should be the first line?

HDembinski:
It should be your name and my name. I am working on a large patch for collector, please don't make any other changes to the collector at the moment.

HDembinski:
@henryiii The collector accumulator is finally moving forward and the implementation seems to be nearly ready. I would like to quickly see it in the boost-histogram Python package, too, since I want to use it in a Python study. What is your time budget right now, do you think you can look into this? Otherwise I can give it a try.

HDembinski:
@wiso I think the PR is almost ready to be merged. I added the copyright notices.

Design changes and additions: I changed the template arguments for the collector, because your original interface was not DRY. I added missing serialization support. You were exposing the internal container, now the internal container remains hidden. Instead, I added basic methods to the collector, which make it iterable and allow element access, and give access to the memory buffer (this is needed by the Python boost-histogram wrapper).

Tests are in place for the interface. I have added a test which checks that the collector template can work with a dynamic array container type as outlined above. 

Coverage is not 100% right now, but that will be fixed in future PRs. The tracking of line coverage in coveralls is a bit broken and I don't know yet how to fix this. I cannot bring coverage to 100 % if the tracking does not work correctly.

Do you wish to add anything at this point?

HDembinski (approved):

wiso:
Hello @HDembinski , thank you very much for your improvement. I am not planning to add anything. I would be happy to see this included.
About the Python package, I have not fully understood how it has been implemented, I thought it was a binding of boost-histogram, but I see several parts are rewritten. I don't have expertise on it, even if it is the main way I use hist.

Issue #392, opened 2023-12-12
CMake Warning from boostorg super project cmake


mborland:
When running cmake configuration from boost root the following warning is issued:

```
CMake Warning (dev) at libs/histogram/test/CMakeLists.txt:5 (find_package):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
  are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.
```

Full build log:
```
-- The CXX compiler identification is AppleClang 15.0.0.15000100
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Boost: using system layout: include, bin, lib, lib/cmake
-- Boost: Debug build, static libraries, MPI OFF, Python OFF, testing 1
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test BOOST_ATOMIC_TARGET_X86
-- Performing Test BOOST_ATOMIC_TARGET_X86 - Failed
-- The C compiler identification is AppleClang 15.0.0.15000100
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Boost.Context: architecture arm64, binary format mach-o, ABI aapcs, assembler gas, suffix .S, implementation fcontext
-- The ASM compiler identification is AppleClang
-- Found assembler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Boost.Fiber: NUMA target OS is none
-- Performing Test BOOST_FILESYSTEM_HAS_INIT_PRIORITY
-- Performing Test BOOST_FILESYSTEM_HAS_INIT_PRIORITY - Success
-- Performing Test BOOST_FILESYSTEM_HAS_CXX20_ATOMIC_REF
[...]

PR #393, opened 2023-12-12
Replace deprecated find_package for python

mborland:
Closes: #392

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Distributed under the Boost Software License, Version 1.0.
 # See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
 
-find_package(PythonInterp)
+find_package(Python3)
 if (PYTHONINTERP_FOUND)
   # checks that b2 and cmake are in sync
   add_test(NAME runpy-${PROJECT_NAME}_check_build_system COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check_build_system.py)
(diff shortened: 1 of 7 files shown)


test/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Distributed under the Boost Software License, Version 1.0.
 # See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
 
-find_package(PythonInterp)
+find_package(Python3)
 if (PYTHONINTERP_FOUND)
```
henryiii, line 6: This is not the same variable. And this requires CMake 3.15+.

henryiii:
```suggestion
if (Python3_FOUND)
```

This is the new variable.

test/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
...
   # checks that b2 and cmake are in sync
   add_test(NAME runpy-${PROJECT_NAME}_check_build_system COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check_build_system.py)
```
henryiii, line 8: You should be able to use the target here, `Python3::Interpreter`. If not, `${Python3_EXECUTABLE}` is the new variable name.

HDembinski: `find_package(Python3::Interpreter)` did not work, but `find_package(Python3 COMPONENTS Interpreter)` does. I will implement that in another minor PR.

henryiii: That’s the default so it’s fine either way, what I meant was `${Python3_EXECUTABLE}` (the variable) can be replaced by `Python3::Interpreter` (a CMake target).

henryiii: (After COMMAND a target is supported)

HDembinski: Ok, I see.

PR #398, opened 2024-04-24
Fix python3 variables

mborland:
Fixes the variables found by `find_package(Python3)` as pointed out by @henryiii


HDembinski:
Weird, why did CI miss this?

HDembinski:
Thanks.

PR #401, opened 2024-07-20
Add support for modular build structure.

grafikrobot:
This is part of the effort to make the Boost libraries "modular" for build and consumption. See https://lists.boost.org/Archives/boost/2024/01/255704.php and https://github.com/grafikrobot/boost-b2-modular/blob/b2-modular/README.adoc for more information.

This PR depends on the following other PRs being merged to both develop and master branches of the respective repos:

- https://github.com/boostorg/boost/pull/854

This PR will be changed to ready for review, i.e. not draft, when the above are merged. Do not merge this one until that time.


HDembinski:
I didn't see this, because I am not getting notifications unless someone mentions me with @HDembinski. I changed something in the options of this repo so that hopefully that doesn't happen again.

@grafikrobot Thank you for the changes to the build. There seems to be an issue with boost.random in the slow tests, a linker error. Do you have an idea how to fix this? You also remove some extra warning flags `cxxflags=-Wno-unused-parameter cxxflags=-Wno-uninitialized`, why?

HDembinski:
> @HDembinski
> 
> > There seems to be an issue with boost.random in the slow tests, a linker error. Do you have an idea how to fix this?
> 
> Have those slow tests ever passed? Asking as I don't see them in previous commits.

Yes, they used to pass. I only run them on PRs, but not on pushes, that's why they don't show up in the list of commits.

HDembinski:
@grafikrobot Thank you for the work, I will look into this on the weekend. The problem for me also is that I am not developing on a Mac anymore, I have a Windows machine now (urgh). I think I can fix all these three issues.

HDembinski:
So, I finally had a closer look.

>     * The coverage fails after doing the building. I'm not familiar with the coverage it's doing to decipher why the lcov file is missing.

It looks like `../../b2 $B2_OPTS toolset=gcc-${GCC_VERSION} cxxstd=latest coverage=on test//all` doesn't compile anything. No tests are run and no coverage data is created. Any idea why? The alias test//all should run all tests.

If I remember correctly, lcov only works with a matching version of gcc. If you know a better way to measure coverage in C++, I'd love to switch to another tool. You upgraded from gcc-11 to gcc-15, I will probably need to upgrade lcov as well.

>     * The slow clang build detects ODR violations.

That's because of boost-serialization if you look closely, which you added as a constant dependency to this build. boost-serialization causes lots of issues, that's why I keep it as an optional dependency that is only compiled for some tests that are not as strict, as the ones I use to check boost-histogram.
 
>     * The slow gcc test detects dangling temp reference error/warning in the main test.

That's a newly discovered bug - it seems.

HDembinski:
@grafikrobot Progress.

The coverage test still fails, because it doesn't compile anything. test//all works in other circumstances. Weird.

The gcc slow test still fails because of that "possibly dangling pointer". As far as I understand, the pointer shouldn't be dangling because of NRVO. The issue is not detected by another other compiler and not caught by the UBSAN / ASAN tests, so perhaps a false positive. As a workaround, we could turn off that warning for gcc for that line.

CMakeLists.txt
```diff
@@ -16,10 +16,10 @@ target_link_libraries(boost_histogram
   INTERFACE
     Boost::config
     Boost::core
+    Boost::math
```
grafikrobot, line 19: Should you remove that to mirror the b2 side where math is optional?

HDembinski: Maybe, but it does cause so many problems, because it is header only.

grafikrobot: I failed to parse that sentence. :-( Just saying the it's preferred to be consistent between the two build systems.

HDembinski: Sorry, forgot a negation there.

boost-serialization is a pain because it is a compiled library. If it is a global dependency, it always gets compiled even if it is not even used in the tests that are actually run. boost-math is header-only, so if the tests that use it are not compiled, it does not cause any issues.

HDembinski: I agree with your point regarding consistency.

HDembinski:
Bizarr. I haven't tried to reproduce locally, but the same issue should show up then if I use the same args. Will try this.

PR #412, opened 2025-06-05
add new docs

### LINKED ISSUE #402: Filling a weighted dynamic rank histogram with a dynamic input
I might have missed something, but I can't figure out a clean way to fill a dynamic rank histogram with inputs that also dynamic without using `fill` and some slightly weird reformatting for the inputs.

Specifically I have something like this

```C++
  using dax_t = bh::axis::regular<double>;
  std::vector<dax_t> axes {
    dax_t(1, -0.5, 0.5, "ax0"),
    dax_t(1, -0.5, 0.5, "ax1")
  };
  auto hdyn = bh::make_weighted_histogram(axes);
  // seems we need to do some funny organization to make weighting
  // work with dynamic axes.
  std::vector<std::vector<double>> vals { {0}, {0} };
  hdyn.fill(vals, bh::weight(0.5));
```
If I try to fill with a `std::vector<double> vals {0, 0};` and using `hdyn(vals, bh::weight(0.5));` I get a `number of arguments != histogram rank` exception. Of course calling `hdyn(0, 0, bh::weight(0.5))` works fine, but that sort of limits the dynamic part of the dynamic axes.

To be clear, what I have works fine and everything else about this library is amazing. And I don't have much of a use case for dynamic axes anyway, so this is more of a curiosity.

### LINKED ISSUE #406: Could the axis method `index` be templated?
I have seen that boost::histogram will not accept a class for an axis where `index` is a template method. I considered that it would be helpful to have the parameter `x` be a template type for certain operations. Is there a strong reason not to allow this?

HDembinski:
Closes #402
Closes #406

This clarifies how to use `histogram::fill` efficiently with multi-dimensional histograms, and discusses a workaround for the design choice that `Axis::index` cannot be overloaded.

And since I haven't build the documentation locally for a long time, and a new computer where I needed to set things up again from scratch, I also added a README.md on how to do that (which is not trivial for boostbook).

### DIFF
--- a/doc/concepts/Axis.qbk
+++ b/doc/concepts/Axis.qbk
@@ -9,7 +9,7 @@
 
 An [*Axis] maps input values to indices. It holds state specific to that axis, like the number of bins and any metadata. Must be [@https://en.cppreference.com/w/cpp/named_req/CopyConstructible CopyConstructible], [@https://en.cppreference.com/w/cpp/named_req/CopyAssignable CopyAssignable], and *nothrow* [@https://en.cppreference.com/w/cpp/named_req/MoveAssignable MoveAssignable].
 
-*Note*: `Axis::index` cannot be templated or overloaded. This is a limitation of the internal code used to detect generic user-defined axes. A workaround for making an axis that accepts multiple value types is described in the [link histogram.guide.expert.axis_multiple_value_types [*Guide]].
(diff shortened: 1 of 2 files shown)


sdarwin:
@HDembinski , please see new pull request. Related to this one.

PR #413, opened 2025-06-06
Docs build with CRLF

sdarwin:
Boost superproject docs builds are failing:  

https://app.circleci.com/pipelines/github/boostorg/boost?branch=develop

cc. @pdimov

### DIFF
--- a/doc/Jamfile
+++ b/doc/Jamfile
@@ -41,7 +41,14 @@ doxygen reference
 
 actions doxygen-postprocessing
 {
-  $(THIS_PATH)/doxygen_postprocessing.py "$(>)"
+  if [ os.name ] = NT
+  {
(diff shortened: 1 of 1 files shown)


HDembinski:
There has to be a better way than reverting. The old code doesn't work on platforms which do not create the python alias, e.g. Ubuntu.

The proper solution here is to get the interpreter from the Python module, but it doesn't expose the interpreter after detecting it AFAIU.

HDembinski:
By the way the boost book docs on the website all contain faulty characters, they used to look correct in the past.

pdimov:
I remember having issues with `python` on containers and switching to `python3`, which seems to work everywhere.

sdarwin:
Updated the PR to check the operating system. It will use `python` on Windows otherwise `python3`.

HDembinski:
Thanks, looks good!

PR #414, opened 2025-06-12
Update doc Jamfile

sdarwin:
Hi Hans,
In almost all cases, I test small code changes in spite of them visually appearing to be obviously correct.   This time seemed like a 1-in-100 chance to skip the rule.  The code "looked right".   And guess what happened! It only reinforces the need for that rule.   So, if you have a chance, please build docs with this new Jamfile.  Or, either way, we need to merge this to fix the build.

### DIFF
--- a/doc/Jamfile
+++ b/doc/Jamfile
@@ -39,17 +39,20 @@ doxygen reference
                              \"BOOST_ATTRIBUTE_NODISCARD\""
 ;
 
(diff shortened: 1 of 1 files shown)


HDembinski:
The real issue: there is no CI test for building docs. I am currently working on rectifying this. Will merge this patch...

PR #418, opened 2025-11-26
Only use console window size if `ioctl(TIOCGWINSZ)` succeeded.

Lastique:
This avoids potentially using uninitialized data if `ioctl` fails. Also, zero-initialize the `winsize` struct to begin with.


Lastique:
This fixes valgrind errors in `histogram_ostream_ascii_test`, when the test output is redirected to a file. Full valgrind log is below:

[vg.log](https://github.com/user-attachments/files/23775283/vg.log)

HDembinski:
Thanks for investigating and providing a solution, that's awesome.

PR #419, opened 2026-01-22
Remove dependencies on Boost.StaticAssert

Lastique:
Boost.StaticAssert has been merged into Boost.Config, so remove the dependency.


henryiii:
Nice. Ping @HDembinski.

HDembinski:
Thanks for the ping @henryiii

Issue #416, opened 2025-08-15
Sparse storage with accumulators


henryiii:
Currently, sparse storage works:

```cpp
#include <boost/histogram.hpp>
#include <boost/histogram/ostream.hpp>
#include <unordered_map>
#include <iostream>
#include <vector>

int main() {
    namespace bh = boost::histogram;
    auto h = bh::make_histogram_with(std::unordered_map<std::size_t, double>(), bh::axis::regular<>(10, 0.0, 1.0));
    std::vector<double> values = {.2, .3};
    std::vector<double> weights = {1, 1};
    h.fill(values, bh::weight(weights));
    std::cout << bh::algorithm::sum(h) << std::endl;
}
```

As does accumulators with dense storage:

```cpp
#include <boost/histogram.hpp>
#include <boost/histogram/ostream.hpp>
#include <iostream>
#include <vector>

int main() {
    namespace bh = boost::histogram;
    auto h = bh::make_histogram_with(std::vector<bh::accumulators::weighted_sum<double>>(), bh::axis::regular<>(10, 0.0, 1.0));
    std::vector<double> values = {.2, .3};
    std::vector<double> weights = {1, 1};
    h.fill(values, bh::weight(weights));
    std::cout << bh::algorithm::sum(h) << std::endl;
}
```

But you can't use the two together:

```cpp
#include <boost/histogram.hpp>
#include <boost/histogram/ostream.hpp>
#include <unordered_map>
#include <iostream>
#include <vector>

int main() {
    namespace bh = boost::histogram;
    auto h = bh::make_histogram_with(std::unordered_map<std::size_t, bh::accumulators::weighted_sum<double>>(), bh::axis::regular<>(10, 0.0, 1.0));
    std::vector<double> values = {.2, .3};
    std::vector<double> weights = {1, 1};
    h.fill(values, bh::weight(weights));
    std::cout << bh::algorithm::sum(h) << std::endl;
}
```

Produces:

<details><summary>Error message (truncated):</summary>
<p>

```
In file included from /opt/wandbox/gcc-13.2.0/include/c++/13.2.0/ext/alloc_traits.h:34,
                 from /opt/wandbox/gcc-13.2.0/include/c++/13.2.0/bits/stl_uninitialized.h:64,
                 from /opt/wandbox/gcc-13.2.0/include/c++/13.2.0/vector:65,
                 from /opt/wandbox/boost-1.83.0-gcc-13.2.0/include/boost/histogram/fwd.hpp:19,
[...]

PR #421, opened 2026-03-25
fix: support sparse storage with accumulators

henryiii:
This is a fix for #416. Added test fails without the fix, compiles and passes with the fix. This was blocking sparse storage in boost-histogram (can't promise it is the only blocker, but was the first one we hit).

Use copilot in VSCode, which selected GPT-5.3-Codex, to convert the issue into a fix and test.


henryiii:
C++17 version, FWIW:

```cpp
auto [it, _] = map->try_emplace(idx, value_type{});
it->second += u;
```

HDembinski:
Looks good, thanks.

henryiii:
Will this make it into 1.91? The deadline for fixes is April 1. (Doesn't have to)

HDembinski:
Yeah, I think I can merge it into the release, because it is a bug-fix.

henryiii:
@HDembinski Will this make it into 1.92? The deadline for all changes is August 5.

(I've got a few other fixes open too, and I can submit a fix for #430 if you want me to).

HDembinski:
yeah let's aim for that

PR #422, opened 2026-04-22
feat: support passing a coverage parameter to project

henryiii:
This would be useful for boost-histogram, as we could support `flow=False` there directly with this. This uses the existing coverage tag, allowing it to be passed in.

:robot: Assisted-by: Copilot:Kimi-K2.6 (an open-source model)

### DIFF
--- a/include/boost/histogram/algorithm/project.hpp
+++ b/include/boost/histogram/algorithm/project.hpp
@@ -33,9 +33,15 @@ namespace algorithm {
   Arguments are the source histogram and compile-time numbers, the remaining indices of
   the axes. Returns a new histogram which only contains the subset of axes. The source
   histogram is summed over the removed axes.
+
+  @param h source histogram.
+  @param cov whether to include underflow/overflow bins in the sum.
+  @param n remaining axis index.
+  @param ns remaining axis indices.
 */
 template <class A, class S, unsigned N, typename... Ns>
-auto project(const histogram<A, S>& h, std::integral_constant<unsigned, N>, Ns...) {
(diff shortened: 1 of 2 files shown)


henryiii:
@HDembinski how does this look?

HDembinski:
project is wrong when you don't have flow bins. Using project without full coverage of flow bins gives wrong results. What's the use case for that? That's why it is intentionally not possible with the API.

Scott Meyers: "Make interfaces easy to use correctly and hard or impossible to use incorrectly."

henryiii:
Because you need to remove them sometimes, and boost-histogram has to work around this by making an intermediate histogram with crop then run projection; having this natively will cut the memory of an operation like `h[0:len:sum]` in half. Chunks of code in the most complex part of `_handle_slice` go away.

This was added to boost-histogram's `.project` as well after a request dating back to 2023. If you make a cut, such as `h_new = h_old[10:50, :, :, :]`, you might want to plot a variable with that cut using `h_new.project(2, flow=False)`. Without the ability to remove flow bins, you can't see the affect of the cut, and you don't want to remove flow bins on the histogram with `h_new_plot = h_old[10j:50j:sum, :, :, :]` (remember it takes double memory to do so, too!) since you want to keep them for later operations. See https://github.com/scikit-hep/boost-histogram/issues/1032.

You should make the interface correct by default (flow=True), but supporting the full range of options is beneficial, especially for plotting.

Also, having the option tells users, via the interface, that it defaults to full coverage. Not having the option at all makes it ambiguous, it's not part of the interface. (I think `h_old[10:50, :].project(x).plot()` becomes less confusing if there's a `flow=True` default in the signature).

PR #431, opened 2026-07-14
Fix scalar-broadcast fill after a non-inclusive axis (scikit-hep/boost-histogram#960)

NJManganelli:
### Problem

In the vectorized fill (`detail/fill_n.hpp`), a scalar argument is broadcast across an
axis by computing that axis' contribution to the linear index once and adding it to
every entry. The contribution was derived from the *change* of the first index
(`*begin_`). When a previous axis had already invalidated some entries — including the
first one — the broadcast axis then invalidated **all** entries and silently dropped
valid data.

This surfaced downstream as scikit-hep/boost-histogram#960
("Histogram with IntCategory and StrCategory axes can be zeroed by out-of-bounds fill
values"): a non-growing category axis combined with out-of-bounds entries and a
broadcast scalar on a second axis produced a sum of zero. One-at-a-time filling was
always correct; only the array/`fill_n` path was affected.

### Fix

Compute the axis contribution on a *fresh* index and add it to all entries, leaving
already-invalid entries untouched. The growing and non-growing scalar paths now share
one code path through `linearize` / `linearize_growth`, so the "delta from `*begin_`"
heuristic is gone entirely.

Regression tests compare the vectorized fill against the one-at-a-time reference for
scalar-broadcast-after-non-inclusive-axis, including the growing-axis, zero-point-shift,
and underflow variants (static and dynamic histograms).

### Note

The reported bug is in the Python bindings, but the defect is here in core `fill_n`;
consuming it there is just a submodule bump. Happy to open the companion PR against
scikit-hep/boost-histogram once this lands.

🤖 Root-caused and drafted with Claude Code; reviewed by @NJManganelli.

### DIFF
--- a/include/boost/histogram/detail/fill_n.hpp
+++ b/include/boost/histogram/detail/fill_n.hpp
@@ -103,19 +103,37 @@ struct index_visitor {
   void call_1(std::true_type, const T& value) const {
     // T is compatible value; fill single value N times
 
-    // Optimization: We call call_2 only once and then add the index shift onto the
-    // whole array of indices, because it is always the same. This also works if the
-    // axis grows during this operation. There are no shifts to apply if the zero-point
-    // changes.
-    const auto before = *begin_;
-    call_2(IsGrowing{}, begin_, value);
-    if (is_valid(*begin_)) {
-      // since index can be std::size_t or optional_index, must do conversion here
-      const auto delta =
-          static_cast<std::intptr_t>(*begin_) - static_cast<std::intptr_t>(before);
-      for (auto it = begin_ + 1; it != begin_ + size_; ++it) *it += delta;
-    } else
+    // Compute the contribution of this axis to the linear index once on a
+    // fresh index and add it to the whole array of indices. The contribution
+    // must not be derived from the change of *begin_, because a previous axis
+    // may have already invalidated some entries, including the first one
+    // (scikit-hep/boost-histogram#960). Entries that are already invalid stay
+    // invalid, all other entries receive the same contribution.
+    //
+    // The fresh index is biased by stride_, because the contribution of an
+    // underflow bin is -stride_ and the index type is unsigned. The bias is
+    // subtracted again when the delta is computed.
+    index_type idx{stride_};
+    if (IsGrowing::value) {
+      // IsGrowing::value is a compile-time constant, the dead branch is
+      // eliminated; linearize_growth also handles non-growing axes.
+      axis::index_type shift;
+      linearize_growth(idx, shift, stride_, axis_,
+                       try_cast<value_type, std::invalid_argument>(value));
+      // No index shifts to apply if the zero-point changes, since all entries
+      // receive the same contribution which is computed after the growth, but
+      // the shift must be recorded so that the storage is resized correctly.
+      if (shift > 0) *shift_ += shift;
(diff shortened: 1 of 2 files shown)


NJManganelli:
I asked Claude to match the comment style of the repo, I don't know that it really did (and it left a comment I asked it to remove regarding a now-gone commit that templated a true/false growable axis to no real gain). I'll be ready to have Claude update based on any adversarial review, human feedback, etc.

include/boost/histogram/detail/fill_n.hpp
```diff
@@ -103,19 +103,37 @@ struct index_visitor {
...
+                       try_cast<value_type, std::invalid_argument>(value));
+      // No index shifts to apply if the zero-point changes, since all entries
+      // receive the same contribution which is computed after the growth, but
+      // the shift must be recorded so that the storage is resized correctly.
+      if (shift > 0) *shift_ += shift;
+    } else {
+      linearize(idx, stride_, axis_, try_cast<value_type, std::invalid_argument>(value));
+    }
+    if (is_valid(idx)) {
+      const auto delta = static_cast<std::intptr_t>(static_cast<std::size_t>(idx)) -
```
HDembinski, line 131: Is this relying on two's complement of the signed -> unsigned -> signed cast or just a superfluous cast? I need another comment here that explains that this double cast is not superfluous but intentional.

include/boost/histogram/detail/fill_n.hpp
```diff
@@ -103,19 +103,37 @@ struct index_visitor {
...
+    // must not be derived from the change of *begin_, because a previous axis
+    // may have already invalidated some entries, including the first one
+    // (scikit-hep/boost-histogram#960). Entries that are already invalid stay
+    // invalid, all other entries receive the same contribution.
+    //
+    // The fresh index is biased by stride_, because the contribution of an
+    // underflow bin is -stride_ and the index type is unsigned. The bias is
+    // subtracted again when the delta is computed.
+    index_type idx{stride_};
+    if (IsGrowing::value) {
```
HDembinski, line 117: looks like this replaces call_2 completely, in that case, those call_2 functions should be removed.

HDembinski: Likewise, call_1 then doesn't need to be called call_1 anymore, just call_impl.

HDembinski:
@NJManganelli Looks good, please resolve the comments, though. The AI tends to not clean up after itself, always afraid to remove things. However, all API inside the detail namespace is fair game, even if public.

henryiii:
Btw, Claude Code added a feature about three days ago `/simplify`, which runs four agents looking for four classes of simplifications (reuse, simplification, efficiency, altitude), and it works really well. Removing API and comments still are two of the hardest things for AI to do, though. :) If you edit your AGENTS.md (CLAUDE.md should be symlink) to state "the detail module is private, API changes are safe inside it" or something like that, it might be a bit better.

Another thing I like is the prompt: "You are an adversarial reviewer for the new feature in this branch. Do you see any problems? Anything that could be done or written more cleanly? Can you break it?".

Anyway, this fixes #426.

HDembinski:
The [ponytail skill](https://github.com/DietrichGebert/ponytail) is also good.

PR #432, opened 2026-07-23
fix: undefined behavior in indexed() for rank-0 histograms

### LINKED ISSUE #430: `indexed()` is undefined behavior for rank-0 (zero-axis) histograms
If you send up with a rank-0 histogram, while some operations support it, some operations can segfault. I ran into this in boost-histogram. It only segfaults on windows. I've got a fix at https://github.com/scikit-hep/boost-histogram/pull/1153, but would be nice to natively support it too. I asked Claude to generate a report (below). I'm on a Mac so didn't do the "before you file" suggestion at the end - you can see the failures in both integration tests CI https://github.com/scikit-hep/integration-tests/actions/runs/27257489976/job/80494999483 and its own: https://github.com/scikit-hep/boost-histogram/actions/runs/27492667553, assume that is enough.

## Summary

Iterating a **rank-0** histogram (zero axes, one cell) with `indexed()` — directly, or indirectly via `algorithm::sum(h, coverage::inner)` / `algorithm::empty(h, ...)` — reads **uninitialized per-axis state** in `indexed_range::iterator::operator++`. The read isn't flagged by ASan/UBSan (uninitialized *stack* read), so the loop usually terminates by luck and returns the correct result. With an unlucky stack layout the uninitialized `end` / `end_skip` values drive `operator++` to advance the underlying iterator past `end()` (or walk `c` off the `indices_` buffer), producing an out-of-bounds read and a crash.

We hit this as a hard crash in the Python bindings ([scikit-hep/boost-histogram#1153](https://github.com/scikit-hep/boost-histogram/pull/1153)): `bh.Histogram()` (rank-0) `.sum()` reproducibly segfaulted on **free-threaded Windows (CPython 3.14t)** while passing on every other platform/build — purely because that build's stack frames differed.

Affected: Boost 1.90/1.91 (develop `5e9b7b1`); the relevant code is unchanged on `develop`.

## Minimal reproducer

```cpp
#include <boost/histogram.hpp>
#include <boost/histogram/algorithm/sum.hpp>
#include <vector>

int main() {
    // Rank-0 requires a *dynamic* axis container with an empty vector.
    using axes_t = std::vector<bh::axis::regular<>>;
    auto h = bh::make_histogram(axes_t{});       // rank() == 0, size() == 1

    bh::algorithm::sum(h, bh::coverage::all);     // OK: uses plain hist iteration
    bh::algorithm::sum(h, bh::coverage::inner);   // UB: drives indexed() over rank-0
}
```

Compiles and (usually) runs cleanly, but `sum(..., inner)` reads uninitialized memory. MemorySanitizer flags it; ASan/UBSan do not. Stability depends entirely on stack contents.

## Root cause

For rank > 0, the `indexed_range` constructor initializes one `index_data` per axis inside a `for_each_axis` lambda (`indexed.hpp:328`). For **rank 0**:

- `for_each_axis` runs **zero times**, so `indices_[0]` (`idx`, `begin`, `end`, `begin_skip`, `end_skip`) is **never initialized** — yet the backing slot exists in the `std::array<index_data, buffer_size>` and is read by `operator++`.
- The empty-range guard `if (begin_ == end_) return;` (`indexed.hpp:323`) does **not** fire: a rank-0 histogram still has one cell, so `begin_ != end_`.

The single iteration then runs `operator++` (`indexed.hpp:251–268`), reading the uninitialized `idx`/`end`/`end_skip`. Because `indices_.end() == indices_.begin()` for rank 0, the `if (++c == indices_.end())` guard never matches once `c` is incremented, so the loop walks `c` forward through uninitialized records while `iter_ += c->end_skip` moves the underlying iterator by arbitrary amounts → OOB read.

## Suggested upstream fixes

1. Extend the early-out to cover rank 0 (treat it as a degenerate single-cell range, stepping `operator++` straight to `end_` without touching `indices_`).
2. In `operator++`, short-circuit when `indices_.begin() == indices_.end()` (rank 0): just `++iter_`.
3. Have `algorithm::sum`/`empty` route rank-0 through the non-`indexed` path (as `sum` already does for `coverage::all`).

## Downstream workaround (already applied)

scikit-hep/boost-histogram routes rank-0 around `indexed()`: `sum()` uses `coverage::all`, `empty()` checks the single cell directly. Rank-0 has no flow bins, so `inner == all`.

---

**One caveat to flag before you file:** I verified the root-cause analysis against the code and confirmed the path is stack-layout dependent (zero-init and uniform pattern-init both happen to terminate cleanly here), but I could **not** produce a portable always-crashing reproducer from macOS — that needs MemorySanitizer, which Apple clang doesn't ship. If you want a hard demonstration to include, the surest route is running the reproducer under MSan (a Linux clang build), or just cite the concrete downstream crash (free-threaded Windows). The maintainer (Hans Dembinski) will likely recognize it from the code analysis alone. Want me to also draft the MSan invocation, or adapt this into the exact GitHub issue body and open it via `gh`?

henryiii:

Fixes #430.

For a rank-0 histogram, `indexed_range::iterator::operator++` dereferenced the first slot of the per-axis index buffer, which is empty when there are no axes. That read uninitialized stack memory and, with unlucky stack contents, walked the cell iterator out of bounds — observed as a reproducible segfault on free-threaded Windows via the Python bindings (scikit-hep/boost-histogram#1153).

The fix skips the index update when the index buffer is empty; the single cell of a rank-0 histogram is still visited exactly once, so `indexed()`, `algorithm::sum(h, coverage::inner)`, and `algorithm::empty` now all work for rank 0.

Regression tests cover `indexed()` (both coverages), `sum`, and `empty` on a rank-0 histogram. They pass only by luck without the fix (the uninitialized read is stack-layout dependent and needs MemorySanitizer to detect reliably), but they pin down the intended behavior.

### DIFF
--- a/include/boost/histogram/indexed.hpp
+++ b/include/boost/histogram/indexed.hpp
@@ -253,6 +253,8 @@ public:
       const auto cbeg = indices_.begin();
       auto c = cbeg;
       ++iter_;
+      // rank-0 histogram has a single cell and no indices to update
+      if (c == indices_.end()) return *this;
       ++c->idx;
       if (c->idx < c->end) return *this;
       while (c->idx == c->end) {
--- a/test/algorithm_empty_test.cpp
+++ b/test/algorithm_empty_test.cpp
@@ -4,6 +4,7 @@
 // (See accompanying file LICENSE_1_0.txt
 // or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+#include <algorithm>
 #include <array>
 #include <boost/core/lightweight_test.hpp>
 #include <boost/histogram/accumulators/weighted_mean.hpp>
@@ -60,6 +61,16 @@ void run_tests() {
 int main() {
   run_tests<static_tag>();
   run_tests<dynamic_tag>();
+
+  {
+    // regression test for https://github.com/boostorg/histogram/issues/430
+    auto h = make(dynamic_tag());
+    BOOST_TEST(empty(h, coverage::all));
+    BOOST_TEST(empty(h, coverage::inner));
+    std::fill(h.begin(), h.end(), 1);
+    BOOST_TEST(!empty(h, coverage::all));
+    BOOST_TEST(!empty(h, coverage::inner));
+  }
 
   return boost::report_errors();
 }
--- a/test/algorithm_sum_test.cpp
+++ b/test/algorithm_sum_test.cpp
@@ -75,5 +75,13 @@ int main() {
   run_tests<static_tag>();
   run_tests<dynamic_tag>();
 
+  {
+    // regression test for https://github.com/boostorg/histogram/issues/430
+    auto h = make(dynamic_tag());
+    std::fill(h.begin(), h.end(), 3);
+    BOOST_TEST_EQ(sum(h), 3);
+    BOOST_TEST_EQ(sum(h, coverage::inner), 3);
+  }
+
   return boost::report_errors();
 }
--- a/test/indexed_test.cpp
+++ b/test/indexed_test.cpp
@@ -210,6 +210,29 @@ void run_comparison_tests(Tag) {
   BOOST_TEST(1 >= a);
 }
 
+// regression test for https://github.com/boostorg/histogram/issues/430
+void run_0d_tests(coverage cov) {
+  auto h = make(dynamic_tag());
+  BOOST_TEST_EQ(h.rank(), 0u);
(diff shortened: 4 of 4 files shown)


HDembinski:
This needs an update now.

I am surprised that a rank 0 histogram still has a single accumulator. I suppose that's consistent with numpy, but not something I remember defining as an explicit edge case.

Issue #437, opened 2026-07-23
Fable review of Boost.Histogram


henryiii:
I did a sweep with Fable for issues. It independency found two existing issues that already have pending fixes, along with some others. My intent is to slowly fix these a few at a time, unless you'd rather something else. Is there anything you'd like me to prioritize or skip?

:robot: _AI text below_ :robot:

All verification is done. Here's the consolidated review.

# Boost.Histogram project review

Seven parallel review agents swept the codebase (core, axes, detail, accumulators, algorithms/fill, performance, modernization); I verified every finding below either by reproduction (ASan/UBSan/compiled repros, sources in the session scratchpad) or by reading the code myself. Two rediscovered known issues are excluded: #426 (broadcast-scalar fill wipes a chunk when a non-inclusive axis drops the first entry — found independently by two agents) and #430 (rank-0 `indexed()`).

## Bugs — memory safety (all reproduced under ASan/UBSan)

1. **`multi_index.hpp:94` — stack-buffer overflow in dynamic-rank `multi_index`.** The static-size specialization (line 50) throws `std::invalid_argument` on a size mismatch; the dynamic specialization copies into `data_[32]` with no check. `h.at(std::vector<int>(40, 0))` on a dynamic-axes histogram overflows the stack (ASan-confirmed) instead of throwing.

2. **`detail/reduce_command.hpp:50` — stack-buffer overflow in `reduce` with too many positional commands.** The bounds check only guards commands with an explicit axis index. `reduce(h, rebin(2), rebin(2))` on a rank-1 tuple-axes histogram writes past the stack command buffer (ASan-confirmed); the explicit form `rebin(1, 2)` correctly throws.

3. **`storage_adaptor.hpp:85` — buffer overflow in `array_impl` converting constructor.** `operator=` and `reset` both throw `std::length_error` when the source exceeds `max_size()`; the converting constructor doesn't check. Reachable purely through the public API by converting a histogram with more cells than the target `std::array` storage holds (ASan-confirmed).

## Bugs — wrong results (reproduced)

[...]

PR #438, opened 2026-07-23
fix: swapped modf outputs in circular variable::value()

henryiii:
From #437.

`std::modf` returns the fractional part and stores the whole part in its output argument. The circular branch of `variable::value()` used the return value as the bin index `k` and the whole part as the interpolation weight `z` — exactly swapped. Every fractional index, and every integer index on a non-equidistant circular axis with more than two bins, returned a point on the line through the first two edges instead of the correct edge, so `bin(i).lower()/upper()/center()` and the printed representation were wrong.

The error cancels for size-2 axes with integer arguments, which is all the existing test covered. Added a size-3 non-equidistant regression test (fails with 5 wrong values before the fix) and fixed the variable use.

### DIFF
--- a/include/boost/histogram/axis/variable.hpp
+++ b/include/boost/histogram/axis/variable.hpp
@@ -210,8 +210,9 @@ public:
     if (options_type::test(option::circular)) {
       auto shift = std::floor(i / size());
       i -= shift * size();
-      double z;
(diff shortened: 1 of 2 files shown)


HDembinski:
Wow.

Issue #440, opened 2026-07-28
Sparse histogram operation improvements


henryiii:
I'm working on sparse support in boost-histogram, and ran into two issues:

The simple one: `map_impl::reference::operator+=` and `operator-=` does `map->emplace(idx, u)` when the cell is missing, even when `u` is zero. Notably `reference::operator=` right above it (line 153) does handle zero correctly — it erases/skips.

The harder one: there's no sparse implementation for some of these operations, so memory usage spikes to the dense storage size.

Ref https://github.com/scikit-hep/boost-histogram/pull/1139.

PR #441, opened 2026-07-28
fix: do not create map cells when adding or subtracting zero

henryiii:
`map_impl::reference::operator+=` and `operator-=` emplaced a cell for a missing index even when the result stayed as default `value_type{}`. Compute the result in a temporary and emplace only if it is non-default, as `operator/=` already does.

First half of #440, the simple fix.

Assisted-by: ClaudeCode:claude-opus-5


henryiii:
Update: actually fixing this fully isn't bad at all (compared to trying to patch it downstream in boost-histogram, especially!), so I'm going to make an alternative full-fix PR in a moment.

PR #442, opened 2026-07-28
feat: sparse storage add/sub

### LINKED ISSUE #440: Sparse histogram operation improvements

henryiii:
This fixes #440. Closes #441 (could be rebased after that one, if you prefer). This keeps `+=`/`-=` from spiking memory on large sparse histograms.

I wasn't sure about protecting for types that don't have `==`. Everything except `/=` and `==` is already protected, similar to the current code here, while `/=`/`==` don't have that protection (like #441), so they can't be used on types without `==` (but everything else works, including `+=`). The concept claims it requires `==`. I stuck with protecting it, but can remove that if you'd prefer. I could also make `/=` work for `==` types (always insert) as a separate PR if you want parity.

Assisted-by: ClaudeCode:claude-fable-5

### DIFF
--- a/include/boost/histogram/detail/detect.hpp
+++ b/include/boost/histogram/detail/detect.hpp
@@ -85,6 +85,10 @@ BOOST_HISTOGRAM_DETAIL_DETECT(is_map_like, ((typename T::key_type*)nullptr,
                                             (typename T::mapped_type*)nullptr,
                                             std::begin(t), std::end(t)));
 
+// storage exposes the container of non-empty cells (sparse storage)
+BOOST_HISTOGRAM_DETAIL_DETECT(has_node_access,
+                              (std::begin(t.node_access()), std::end(t.node_access())));
+
 // ok: is_axis is false for axis::variant, because T::index is templated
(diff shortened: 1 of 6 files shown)


HDembinski:
I am leaning towards calling this a feat instead of a fix, you are special-casing on node_access. I merged the other one, that was more straight forward.

Not sure whether the release manager will accept so many fixes in the beta phase. This is supposed to be the time for small fixes.

HDembinski:
Please rebase.

henryiii:
Since the other one didn't protect on `==` type, and it's merged, I'll drop that from this one too.

HDembinski:
Very nice, thanks.

